PR #328, opened 2014-07-09
Add `select_by_text` methods

blueyed:
With select option lists, it is more convenient to select them by their
(visible) text.

Let me know if you find this useful, and I will add tests.


andrewsmedina:
@blueyed thank you! sorry for the delay.

PR #341, opened 2014-08-27
In case of StaleElementReferenceException, refresh the browser

dmelamedcl:
Occasionnaly, the function `is_text_present` may throw a `StaleElementReferenceException` due to a DOM change. This PR catches the exception and is issuing a browser refresh.


andrewsmedina:
Reload the page is a little intrusive. We will fix it ignoring the `StaleElementReferenceException`: https://github.com/cobrateam/splinter/pull/396/files

PR #344, opened 2014-10-02
Add selection option by text to the element api

StephenDavidson:
This pull request adds a select option by text to the element api. This resolves an issue wherein you are unable to select an option by text because the value occurs before the current element in a different element. (EX. Two state dropdowns on a page with the same values)


andrewsmedina:

PR #345, opened 2014-10-02
add options argument for chrome driver

StephenDavidson:
Allows user to add custom options to the chrome browser.


andrewsmedina:
@StephenDavidson  thank you. Can you add a test?

neilsh:
@andrewsmedina  What would a test for this actually entail? It seems like this change is necessary for the existing tests to actually run properly.

PR #356, opened 2014-11-17
added ignore_robots flag to zope.testbrowser

maccesch:
Otherwise there was no possibility to tell the mechanize browser that is used by the zope.testbrowser driver to ignore robots.txt


NetstationMurator:
That's a feature we'll looking for. I see that Travis mark it as failed, but I see that it's not related to zope.testbrowser. Can someone confirm that it's not a bug in that commit?

andrewsmedina:
@maccesch thank you!

PR #365, opened 2015-01-18
Fixed grammar issue.

frodopwns:

PR #366, opened 2015-01-18
Same grammar issue as #365

frodopwns:


andrewsmedina:
@frodopwns thank you!

PR #367, opened 2015-01-18
Update to indicate that get_iframe() can be used with other attributes..

frodopwns:
...than name.

The implementation of get_iframe(name) should probably be updated so it doesn't imply that only name can be used.


andrewsmedina:
@frodopwns thank you!

PR #374, opened 2015-02-27
Use a minimum selenium version; easier to update selenium in parent packages.

maparent:
Newer firefox versions often break with older selenium.


andrewsmedina:
@maparent thank you!

Issue #165, opened 2012-08-27
splinter on python 3


andrewsmedina:
For this it is need that lxml, selenium, and zope test browser work in python 3

lrowe:
@fgimian I don't think your Python 3 compatibility update got things quite right. I was using splinter 0.6.0 with webdriver under Python 3 and things worked, but after upgrading to 0.7.1 I am seeing many errors like this:
```
>       assert text in context.browser.title
E       TypeError: Type str doesn't support the buffer API
```

I don't think you should be decoding strings to bytes for the title or html. Instead just leave it as an encoded string:

    @property
    def title(self):
        if sys.version_info[0] > 2:
            return self.driver.title.encode('utf-8')
        else:
            return self.driver.title

https://github.com/cobrateam/splinter/commit/5d14dad0b43a7f396f9a9b2d0601ca27e8d78bfc?diff=unified#diff-57bd95ad52ad7f3fc6217684677d0576R43

If a different driver returns a bytes object then in probably should be decoded to a string.

Liang-WenYan:
Awesome!

PR #380, opened 2015-03-20
Return decoded text strings for title and html.

lrowe:
See discussion on #165 - this essentially takes the inverse approach to #304 so that titles and html source are text strings rather than byte strings which makes writing tests feel much more natural.

### DIFF
--- a/splinter/driver/djangoclient.py
+++ b/splinter/driver/djangoclient.py
@@ -141,17 +141,17 @@ class DjangoClient(DriverAPI):
         try:
             return self._html
         except AttributeError:
-            self._html = lxml.html.fromstring(self.html.decode('utf-8'))
+            self._html = lxml.html.fromstring(self.html)
             return self._html
 
     @property
     def title(self):
         html = self.htmltree
-        return html.xpath('//title')[0].text_content().strip().encode('utf-8')
+        return html.xpath('//title')[0].text_content().strip()
 
     @property
     def html(self):
-        return self._response.content
+        return self._response.content.decode(self._response.charset)
 
     @property
     def url(self):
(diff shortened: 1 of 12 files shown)


andrewsmedina:
@lrowe thank you!

splinter/driver/djangoclient.py
```diff
@@ -141,17 +141,17 @@ def htmltree(self):
...
     @property
     def title(self):
         html = self.htmltree
-        return html.xpath('//title')[0].text_content().strip().encode('utf-8')
+        return html.xpath('//title')[0].text_content().strip()
 
     @property
     def html(self):
-        return self._response.content
+        return self._response.content.decode(self._response.charset)
```
doup: I think there is a regression here, Django 1.7 TemplateResponse doesn't have a "charset" property. This seems to be [fixed](https://github.com/cobrateam/splinter/blob/master/splinter/driver/djangoclient.py#L172) on `master` but makes 0.7.2 unusable with Django 1.7.

lrowe: Sorry, I don't use Django myself so only ran the tests. Perhaps replace it with ``getattr(self._response, 'charset', 'utf-8')``?

doup: No problem, I just wanted to document it so that other 1.7 users know that [0.7.2 fails](https://github.com/cobrateam/splinter/blob/0.7.2/splinter/driver/djangoclient.py#L154). I don't think it's necesary to change anything as it seems that it's already [fixed in master](https://github.com/cobrateam/splinter/blob/master/splinter/driver/djangoclient.py#L172):

    return self._response.content.decode(self._response._charset or 'utf-8')

This should be working in 0.7.3. :smile:

PR #385, opened 2015-03-28
Support for Django 1.7 and minor fixes

dryobates:


andrewsmedina:
@dryobates thank you!

PR #386, opened 2015-04-04
Support for Django 1.8

izabeera:


splinter/driver/djangoclient.py
```diff
@@ -8,7 +8,8 @@
 import os.path
 import re
 import sys
-import urlparse
+import six
```
andrewsmedina: I believe that to use `six` from `django.utils.six` is a better way.

andrewsmedina: It is not a problem. But it is needed to add the `six` dependencie in setup.py: https://github.com/kidosoft/splinter/blob/master/setup.py#L28

izabeera: Thanks. Done :)

andrewsmedina:
@izabeera thank you!

PR #392, opened 2015-04-23
android driver

sureshvv:
Selenium 2.45 supports this


lrowe:
Why doesn't the existing remote webdriver client work here?

sureshvv:
It works. Mostly. But still, there are a few reasons for this:

1.Recent selenium version has a webdriver called android. We can map splinter directly to that driver.

2. Certain operations are not supported by AndroidDriver (such as setWindowSize). We can handle those better.

Thank you for looking into this. Will be happy to make this mergeable.

andrewsmedina:
@sureshvv sorry for the delay. Can you add tests?

andrewsmedina:

PR #393, opened 2015-04-25
add driver cookies to request handler connection

samuelcolvin:
Current visit doesn't really worked if your working with a system where a login is required as the "connect" function [here](https://github.com/cobrateam/splinter/blob/af0054a526756916f15fd26fee6c45c6c5d02336/splinter/driver/webdriver/__init__.py#L180) doesn't have the same cookies as the driver being used otherwise. This fixes that.

I presume we might need tests for this and some other checks before adding the cookies? Let me know what and where and I'll add.


andrewsmedina:

PR #396, opened 2015-04-29
Check for StaleElementReferenceException

iffy:
I don't know how to test this -- especially since I can't duplicate the problem all the time.  I just know that the monkey-patched version I am running has not yet raised this error and continues to work.

I'm happy to make modifications (including adding tests) with some guidance.

### DIFF
--- a/splinter/driver/webdriver/__init__.py
+++ b/splinter/driver/webdriver/__init__.py
@@ -11,6 +11,7 @@ import sys
 from contextlib import contextmanager
 
 from selenium.common.exceptions import NoSuchElementException
+from selenium.common.exceptions import StaleElementReferenceException
 from selenium.webdriver.common.action_chains import ActionChains
 
 from splinter.driver import DriverAPI, ElementAPI
@@ -299,6 +300,10 @@ class BaseWebDriver(DriverAPI):
                 # This has occasionally been observed. Assume that the
                 # page isn't fully loaded yet
                 pass
+            except StaleElementReferenceException:
+                # This exception is sometimes thrown if the page changes
+                # quickly
+                pass
         return False
 
     def is_text_not_present(self, text, wait_time=None):
@@ -314,6 +319,10 @@ class BaseWebDriver(DriverAPI):
                 # This exception will be thrown if the body tag isn't present
                 # This has occasionally been observed. Assume that the
                 # page isn't fully loaded yet
+                pass
+            except StaleElementReferenceException:
+                # This exception is sometimes thrown if the page changes
+                # quickly
                 pass
         return False
 


andrewsmedina:
@iffy this exceptions is raised when the page is modified. Why ignore it?

iffy:
@andrewsmedina I don't know? :)

Here's my experience: I was using `is_text_present` to determine if a certain page had loaded.  It would sometimes fail with `StaleElementReferenceException` but not always.  Without this patch, I need to change this:

```
if not browser.is_text_present(text):
    # do something
    pass
else:
    # do something else
    pass
```

to something like this code, which pretty much duplicates the guts of `is_text_present`:

```
found = False
endtime = time.time() + browser.wait_time
while time.time() < endtime:
    try:
        if browser.is_text_present(text):
            found = True
            break
    except StaleElementReferenceException:
        pass
if found:
    # do something
    pass
else:
    # do something else
    pass
```

I'm not sure when I would want the `StaleElementReferenceException` to stop the `is_text_present` check (maybe other people have other use cases I can't think of).

Maybe a better solution would be to let the caller choose which exceptions to ignore?  e.g. 

```
browser.is_text_present(text, ignore_exceptions=(StaleElementReferenceException, SomeOtherException))
```

jaraco:
@andrewsmedina Can we expect a merge with this change anytime soon?

jaraco:
I tried creating a workaround by subclassing the browser object:

```
class Browser(splinter.Browser):
	workaround_396 = jaraco.functools.retry(
		cleanup=functools.partial(time.sleep, 1),
		trap=selenium.common.exceptions.StaleElementReferenceException,
		retries=2,
	)

	@workaround_396
	def is_text_present(self, *args, **kwargs):
		super().is_text_present(*args, **kwargs)

	@workaround_396
	def is_text_not_present(self, *args, **kwargs):
		super().is_text_not_present(*args, **kwargs)
```

But because Browser isn't a class (despite the signal in its casing), that fails with `TypeError: function() argument 1 must be code, not str`.

jaraco:
Ultimately, I ended up bypassing the `splinter.Browser` interface and instead patch the intended driver explicitly:

```
class Browser(splinter.browser.ChromeWebDriver):
	workaround_396 = jaraco.functools.retry(
		cleanup=functools.partial(time.sleep, 1),
		trap=selenium.common.exceptions.StaleElementReferenceException,
		retries=2,
	)

	@workaround_396
	def is_text_present(self, *args, **kwargs):
		super().is_text_present(*args, **kwargs)

	@workaround_396
	def is_text_not_present(self, *args, **kwargs):
		super().is_text_not_present(*args, **kwargs)
```

I'm not yet sure this approach works, but I'm hopeful it will.

andrewsmedina:
@iffy thanks for contribute

PR #404, opened 2015-05-24
browser stack driver implemented.

mroohian:
In order for **test_webdriver_browserstack.py** to pass _Browserstack_ credentials must be provided.

### DIFF
--- a/splinter/browser.py
+++ b/splinter/browser.py
@@ -8,6 +8,7 @@ import sys
 
 from splinter.driver.webdriver.firefox import WebDriver as FirefoxWebDriver
 from splinter.driver.webdriver.remote import WebDriver as RemoteWebDriver
+from splinter.driver.webdriver.browserstack import WebDriver as BrowserstackWebDriver
 from splinter.driver.webdriver.chrome import WebDriver as ChromeWebDriver
 from splinter.driver.webdriver.phantomjs import WebDriver as PhantomJSWebDriver
 from splinter.exceptions import DriverNotFoundError
@@ -16,6 +17,7 @@ from splinter.exceptions import DriverNotFoundError
 _DRIVERS = {
     'firefox': FirefoxWebDriver,
     'remote': RemoteWebDriver,
+    'browserstack': BrowserstackWebDriver,
     'chrome': ChromeWebDriver,
     'phantomjs': PhantomJSWebDriver,
 }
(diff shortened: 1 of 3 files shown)


lrowe:
Why is a specific browserstack driver needed? Credentials can be passed as keyword args when instantiating the remote Webdriver class: https://github.com/cobrateam/splinter/blob/0.7.2/splinter/driver/webdriver/remote.py

neilsh:
I'm not the original submitter (and I realize this is 2 years after the PR was first created), but I think one reason is that browserstack's Selenium DesiredCapabilities takes keys with periods in them. [For example](https://www.browserstack.com/automate/capabilities): `browserstack.video`, `browserstack.timezone`, and `browserstack.debug`, and `browserstack.ie.noFlash`. 

Since each key in DesiredCapabilities is derived from the name of an argument passed to the `WebDriver()` constructor, I'm not sure there's any way to pass these keys using the remote WebDriver as it currently stands. As an abbreviated example, the following isn't valid python:

```python
Browser(driver_name='remote', browserstack.debug=True, browserstack.video=True)
```

lrowe:
@nelish It's possible to pass arbitrary strings using keyword argument unpacking:
```python
Browser(driver_name='remote', **{"browserstack.debug": True, "browserstack.video": True})
```

neilsh:
Excellent point about the argument unpacking. In that case, yeah, I can't think of any reason one would need specific browserstack support. ;)

neilsh:
Also, I can confirm that the argument-unpacking solution you outlined works perfectly with browserstack. Here's an example that's worked well for me:

```python
wd_server = 'http://hub-cloud.browserstack.com/wd/hub'
browser = Browser(
                driver_name='remote',
                platform='windows',
                url=wd_server,
                browserName='chrome',
                name='TEST NAME',
                project='PROJECT NAME',  # used for grouping in browserstack UI
                resolution='1024x768',
                **{
                    'browserstack.debug': True,  # whether to take screenshots at each step
                    'browserstack.video': True,  # whether to record video of session
                    'browserstack.user': browserstack_user,
                    'browserstack.key': browserstack_key,
                }
            )
```

andrewsmedina:
I'm -1 for a specific driver for browserstack

neilsh:
Same here, although it might be worthwhile to include an example for browserstack (and other popular cloud services) in the documentation, similar to the example I posted above.

andrewsmedina:
@neilsh I agree with you. Can you send a PR adding your example in the documentation?

andrewsmedina:

PR #410, opened 2015-06-11
Enable cookie support for remote webdriver

lrowe:
I've been monkey patching this in for a while and it works with Chrome running on Sauce Labs at least.


brianlucas:
Thanks for the patch - it works for me.

andrewsmedina:
Thank you @lrowe.

PR #417, opened 2015-06-23
Added selenium capabilities param for firefox driver

magdalene:
I would like to pass some writable capabilities to the wrapped selenium driver. I've only added it for firefox because I didn't really want to try to make a big change as a first-time contributor.


andrewsmedina:
@magdalene thank you!

PR #420, opened 2015-07-01
Fix find_by_text to deal with "

ecesena:


andrewsmedina:
Hi @ecesena. thank's for your contribution.

Can you add a test for this case?

ecesena:
Added - could you double check if the style is ok?
Travis fails on some versions od django, but it seems unrelated to my change.

andrewsmedina:
@ecesena thank you!

PR #421, opened 2015-07-10
SyntaxError

bnordlund:
Python Shell (Python 3.2.3) on Raspberry Pi (Raspbian) was throwing an error on "import splinter". I suspect two typos.


andrewsmedina:
Splinter supports python 3.4+. So I'm closing this.

PR #423, opened 2015-07-29
Improved support for Chrome in Remote driver

squarelover:
Chrome supports cookies, and mouse actions and is also available in the remote driver. Also, a previous commit that enabled cookie support in the remote driver didn't update the corresponding tests. 
This pull request also has a small number of 100ms sleep statements after click events. There seems to be a lot of latency on click events in the Remote driver, and these tiny bits of sleep compensate for the latency. All tests seem to pass now for the Remote driver.

I know sleep statements aren't the best overall, but I figured for tests, it shouldn't be to bad. In the long run, we may want to investigate some latency compensation code in the Remote driver.


andrewsmedina:

PR #425, opened 2015-08-06
:arrow_up: selenium 2.47.1

nikolas:


andrewsmedina:
@nikolas thank you. You should change https://github.com/cobrateam/splinter/blob/master/setup.py#L26 too.

PR #433, opened 2015-09-19
something unclear

objarni:
Please clarify this part of the splinter documentation


andrewsmedina:
@objarni thank you for your feedback. The Makefile that the docs refers is https://github.com/cobrateam/splinter/blob/master/Makefile

Do you think if we add a mention to clone the project, enter in the project dir and then run `make test` will fix this problem?

objarni:
@andrewsmedina yes I think that would be enough, if that is enough to make "make" work :). 

:+1:

douglascamata:
@andrewsmedina what this issue fixed? I'm trying to clean up our PRs :P

Issue #377, opened 2015-03-14
Can't open Firefox that is installed at a custom path


dwt:
Hi there,

I've got my development browsers installed in /Users/$user/Applications/Network/Browser/ - instead of in /Applications.

This however brings the problem that splinter doesn't find my firefox - and as far as I can see from the source code there is also no ability to set a custom path to find Firefox by.

Well, what to do?

For some ruby app I'm working on, I wrote this code to find the firefox binary:

``` ruby
# -*- encoding : utf-8 -*-
# Only needs doing on OS X
if RbConfig::CONFIG["host_os"].include? 'darwin'

  # Find it with  Spotlight
  firefox_paths = %x[mdfind "kMDItemFSName = Firefox.app"]
  firefox_path = firefox_paths.split('\n').first.chomp
  firefox_executable_path = File.join(firefox_path, '/Contents/MacOS/firefox')

  # Another way would be to dump launch services database and grep it out there
  # /System/Library/Frameworks/CoreServices.framework/Versions/A/Frameworks/LaunchServices.framework/Versions/A/Support/lsregister -dump | grep -n7 org.mozilla.firefox | grep path: | awk '{ print($3) }'

  require 'selenium-webdriver'
  Selenium::WebDriver::Firefox::Binary.path=firefox_executable_path

end
```

Please either adopt something based on this code (this will always find firefox regardless where it is installed) or allow the path to be set as an argument to the firefox web driver wrapper. (Similar settings might be necessary for the other browsers).

dwt:
This workaround seems to work for me, maybe you an adapt it to your needs:

```python
class Monkey(object):
    def patch(self):
        from selenium.webdriver.firefox.firefox_binary import FirefoxBinary
        from subprocess import check_output
        import os
        def find_firefox_path(self):
            firefox_path = check_output(["mdfind", "kMDItemFSName = Firefox.app"]).decode('utf-8').strip()
            return os.path.join(firefox_path, "Contents/MacOS/firefox-bin")
        # FIXME: only on mac
        FirefoxBinary._get_firefox_start_cmd = find_firefox_path
Monkey().patch()
```

dwt:
Hello Maintainers - what do you think, could you include this feature into splinter proper? That way splinter would find firefox binaries on mac regardless of where they are installed.

Some customization would still be needed for people with multiple versions of firefox, but it's need should be greatly diminished.

douglascamata:
If you create docs and tests we gladly accept your PR, friend.

[...]

PR #437, opened 2015-10-08
Issue 377 start firefox from custom path

dwt:
Pull request for https://github.com/cobrateam/splinter/issues/377.

I have to say that I'm not particularly proud of the unit tests as they just mock out half the universe. I'd love to get some guidance how you would consider testing this.

### DIFF
--- a/splinter/driver/webdriver/firefox.py
+++ b/splinter/driver/webdriver/firefox.py
@@ -9,6 +9,7 @@ from subprocess import check_output
 
 from selenium.webdriver import DesiredCapabilities, Firefox
 from selenium.webdriver.firefox.firefox_profile import FirefoxProfile
+from selenium.webdriver.firefox.firefox_binary import FirefoxBinary
 from splinter.driver.webdriver import (
     BaseWebDriver, WebDriverElement as WebDriverElement)
 from splinter.driver.webdriver.cookie_manager import CookieManager
@@ -46,23 +47,10 @@ class WebDriver(BaseWebDriver):
                 firefox_profile.add_extension(extension)
         
         firefox_binary = None
-        from selenium.webdriver.firefox.firefox_binary import FirefoxBinary
         if firefox_binary_path is not None:
             firefox_binary = FirefoxBinary(firefox_path=firefox_binary_path)
-        
-        # Selenium guesses really badly for firefox on os x, so let's guess better
-        standard_firefox_path = "/Applications/Firefox.app/Contents/MacOS/firefox-bin"
-        if firefox_binary_path is None \
-                and sys.platform.startswith('darwin') \
-                and not os.path.isfile(standard_firefox_path):
-            path = check_output(["mdfind", "kMDItemFSName = Firefox.app"]).decode('utf-8').strip()
-            if path != '':
-                if len(path.split()) > 1: # found multiple foxes, just take the first one
-                    # if you need more control, specify manually
-                    # TODO consider some logging?
-                    path = path.split()[0]
-                firefox_binary_path = os.path.join(path, "Contents/MacOS/firefox-bin")
-                firefox_binary = FirefoxBinary(firefox_path=firefox_binary_path)
+        else:
+            firefox_binary = self._guess_firefox_path_if_sensible()
         
         self.driver = Firefox(firefox_profile,
                               capabilities=firefox_capabilities,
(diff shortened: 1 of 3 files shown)


dwt:
Ah, damn. I forgot to mention that I'm having quite some trouble getting the full test suite to run on my machine - the Firefox driver specific tests seem to work fine, but I'm getting lots of errors that look quite unrelated to me.

douglascamata:
I will take a look tomorrow at lunch time. Already 00:30am for me. Thanks!

douglascamata:
1. Can you, please, break your code into smaller functions, specially the improved guess for OS X Firefox's path?

2. I would make a normal functional test first for Linux that temporarily removes Firefox from the path and try to create a Browser instance passing that Firefox binary path manually, avoiding mocks as much as possible. Then I would create a mock test for the improved OS X guess, mostly because we can't run this on CI server. With the code broken into smaller functions testing OS X specific guess will be easier, I think. But we still don't have a egg for mocking stuff. For this, IMHO, we can add the `mock` egg, which is the official `unittest` mock solution for Python 3.3 and beyond. What you think, @andrewsmedina?

3. Also, please move `selenium.webdriver.firefox.firefox_binary.FirefoxBinary` import to the top of the Firefox webdriver. 

4. Documentation should describe the behavior when two Firefox binaries are found in OS X

Thank you very much!

dwt:
Hi @douglascamata,

I tried to heed your recommendations, however I encountered some issues doing so.

1. Extracting the guessing method, didn't do too much good, as the constructor WebDriver() already has the unwanted side effect of actually opening the browser, so I still needed the mocks. :(

2. Rewriting the test for the setting of a custom path to extract FF from the path and see that it is then set from the right path is a test that would always fail on anything but linux, as the whole point of this patch is that selenium has quite some autodetection on where to find FF on Linux and Windows - but not on mac. See also: https://github.com/SeleniumHQ/selenium/issues/1128 . So what to do? I'd like to provide a test that I can a) test myself (mac user) and b) works on any platform so other people don't see spurious failures running the test. Also I'd like not to restrict the test to only darwin, so you can run it on the integration server.

3. I haven't yet pulled in mock, as @andrewsmedina hasn't yet said anything - in the meantime
   ```python
from contextlib import contextmanager
@contextmanager
def spy_on(an_object, slot_name, spy):
    original_value = getattr(an_object, slot_name)
    setattr(an_object, slot_name, spy)
    try:
        yield spy
    finally:
        setattr(an_object, slot_name, original_value)
```
isn't too bad I think, and should be quite easy to replace once you decide to pull in mock.

What do you think?

**Comment by dwt** [comment]
Hi friends,

any update on this, did you have time yet to take a look at my changes?

Best regards,
Martin

**Comment by douglascamata** [comment]
@dwt I agree with all your points and read your issue at Selenium (I see did you maybe will have some progress there with a contribution). I would just like to wait for another colleague review, but it seems that @andrewsmedina is very busy. @fsouza, can you do a quick review here for me, please?

**Comment by josemazo** [comment]
Hey guys? How is this new feature going? A soon 0.7.4 release with Firefox path selector will be amazing!

**Comment by dwt** [comment]
I'm currently trying to find the time to port the auto detection part of this patch over to selenium.

I'd love to see this go in to splinter in the meantime as this should be a nice usability boon. As soon as selenium has proper has the auto detection included and released, it could be removed from splinter again, but I guess that will take some time.

**Comment by trojkat** [comment]
Can we skip the auto detection part and merge only the `firefox_binary_path` feature? C'mon guys, let's to this!

**Comment by bmcculley** [comment]
@trojkat it's already possible to set a custom Firefox binary path.

```
from splinter import Browser

binary_path = {'binary_path':'<path/to/binary>'}

with Browser('firefox', **binary_path) as browser:
    browser.visit('http://example.com')
    browser.quit()
```

**Comment by trojkat** [comment]
@bmcculley with this patch yes, but not with current master
https://github.com/cobrateam/splinter/blob/master/splinter/driver/webdriver/firefox.py#L20

**Comment by andrewsmedina** [comment]
@trojkat I will take a time to merge the custom binary path. So @dwt can open another PR with the auto detection.

File: splinter/driver/webdriver/firefox.py [hunk 1]
```diff
@@ -42,9 +45,16 @@ def __init__(self, profile=None, extensions=None, user_agent=None,
         if extensions:
             for extension in extensions:
                 firefox_profile.add_extension(extension)
-
+        
+        firefox_binary = None
+        if firefox_binary_path is not None:
```
amcorreia: This should be:
if firefox_binary_path:
More pythonic way...

splinter/driver/webdriver/firefox.py
```diff
@@ -54,3 +64,23 @@ def __init__(self, profile=None, extensions=None, user_agent=None,
...
+    def _guess_firefox_path_if_sensible(self):
+        # Selenium guesses really badly (i.e. not at all) where Firefox on os x is installed.
+        # So let's ask the system instead
+        standard_firefox_path = "/Applications/Firefox.app/Contents/MacOS/firefox-bin"
+        if not sys.platform.startswith('darwin') \
+            or os.path.isfile(standard_firefox_path):
+            return None
+        
+        path = check_output(["mdfind", "kMDItemFSName = Firefox.app"]).decode('utf-8').strip()
+        if path == '': return None
```
amcorreia: if not path:
  return None

andrewsmedina:

PR #438, opened 2015-10-08
Less memory consumption by leaked firefoxes

dwt:
Well, just something small I noticed where I had lots of open leaked fireboxen around after running some of the tests.

It annoyed me, and this way the tests run way quicker on my machine, due to less ram consumption.

What do you think?


dwt:
Damn, too many commits in this one.

Issue #441, opened 2015-10-17
There should be an option to use html5lib instead of lxml.html in DjangoClient (chokes on some html5 input)


frankier:
It looks like libxml2's html parsing doesn't produce a proper html5 DOM and sometimes chokes on valid html5 even when run in tolerant mode which can result in errors like "XMLSyntaxError: ... Tag footer invalid". The solution is probably to allow the usage of html5lib instead. One hitch with this is the methods from HTMLMixin no longer exist, so the dependence on these should be removed from Splinter.

andrewsmedina:
+1 to use html5lib

adamlwgriffiths:
I've found lxml2's html parser to be unable to handle any real-world HTML.
However, I found html5lib has a habit of closing parent tags off early, causing the children become siblings.
I personally found the inbuilt Python parser superior to html5lib.

PR #442, opened 2015-10-17
Fix repetition in flask and django clients

frankier:
There's a lot of repetition in the Flask and Django clients. In preparation for allowing html5lib to be used in place of libxml2 (https://github.com/cobrateam/splinter/issues/441) I thought these should be merged. I'm a bit confused as to why so much code was just copied and pasted in the first place?


andrewsmedina:
@frankier thank you!

Issue #428, opened 2015-08-18
Multiple select support in Flask driver


twolfson:
During the addition of splinter/Flask driver to one of our projects, we encountered an issue with forms containing select's with the `multiple` attributes. They were serializing as:

```bash
'pets': <MultipleSelectOptions {'cat', 'dog'} for select name='pets'>
```

We currently put together a patch which detects the `MultipleSelectOptions` class and tests with it. However, I am not sure what the best place for this is as it should likely be shared between the Flask and Django clients (or maybe use another approach without a `serialize` method:

https://github.com/cobrateam/splinter/compare/cobrateam:5317b71...underdogio:8b550a0

Let us know what you think it best and we will open up a PR.

twolfson:
Another edit is the `fill_form` currently upcasts values for select as a `list` (regardless of whether this is a list or not):

https://github.com/cobrateam/splinter/compare/cobrateam:5317b71...underdogio:49c91a7#diff-df553efba1487ab0f96b4053dee86c25R256

twolfson:
Any word on this? We are still more than happy to open a PR but we need feedback first.

twolfson:
Went ahead and created a PR anyway -_-;;

PR #443, opened 2015-10-26
Added multi-select support to LxmlDriver

twolfson:
As initially described in #428, we need multi-select support for Flask (i.e. `<select multiple>`). In this PR:

- Added test for serializing a form for a select with multiple values
- Added `serialize` method for testing serialization exclusively
- Added fix for `fill_form` to support accepting multiple values

**Missing:**

- Test for `fill_form` with list coercion. I am not sure why my past self never wrote a test for that


andrewsmedina:
@twolfson thank you!

PR #444, opened 2015-10-26
Added custom headers support to Flask client

twolfson:
In our test suite, we need to send a HTML-specific `Accept` header in order to differentiate HTML requests from JSON ones. In order to add support for that, we need custom headers from `splinter's Flask` client. In this PR:

- Added custom header support for Flask client (based off of Django client)
- Added test

**Notes:**

For anyone who wants `0.7.3` compatible support, we wrote that out here:

https://github.com/underdogio/splinter/compare/6993e85...underdogio:e7b81e5

### DIFF
--- a/splinter/driver/flaskclient.py
+++ b/splinter/driver/flaskclient.py
@@ -57,10 +57,11 @@ class FlaskClient(LxmlDriver):
 
     driver_name = "flask"
 
-    def __init__(self, app, user_agent=None, wait_time=2):
+    def __init__(self, app, user_agent=None, wait_time=2, custom_headers=None):
         app.config['TESTING'] = True
         self._browser = app.test_client()
         self._cookie_manager = CookieManager(self._browser)
+        self._custom_headers = custom_headers if custom_headers else {}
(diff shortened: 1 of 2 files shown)


splinter/driver/flaskclient.py
```diff
@@ -57,10 +57,11 @@ class FlaskClient(LxmlDriver):
 
     driver_name = "flask"
 
-    def __init__(self, app, user_agent=None, wait_time=2):
+    def __init__(self, app, user_agent=None, wait_time=2, custom_headers=None):
```
andrewsmedina: I believe that `headers` is a better name for this parameter.

twolfson: This was mimicked after `DjangoClient` which uses `custom_headers`:

https://github.com/cobrateam/splinter/blob/6993e85e48088c3f6546e2e4be83d52dce863ea5/splinter/driver/djangoclient.py#L65

andrewsmedina: Oh. You're right :)

andrewsmedina:
@twolfson thank you!

PR #445, opened 2015-10-29
Added support for `in` against cookies

twolfson:
During our tests, we wanted to assert some cookies were not present. However, `flaskclient` was erroring out since it was accessing a key that didn't exist via `__getitem__`. As documented by `__contains__`, `__getitem__` is invoked when `__contains__` isn't defined.

https://docs.python.org/2/reference/datamodel.html#object.__contains__

```python
self.assertIn('a', browser.cookies)
# Same as `assert 'a' in browser.cookies
self.assertNotIn('b', browser.cookies)
# Same as `assert 'b' not in browser.cookies
```

To remedy that, we wrote out `__contains__` for each of our engines. In this PR:

- Added `in`/`not in` support for cookies via `__contains__`
- Added test

### DIFF
--- a/splinter/driver/djangoclient.py
+++ b/splinter/driver/djangoclient.py
@@ -49,6 +49,9 @@ class CookieManager(CookieManagerAPI):
     def __getitem__(self, item):
         return self._cookies[item].value
 
+    def __contains__(self, key):
+        return key in self._cookies
+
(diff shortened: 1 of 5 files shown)


douglascamata:
Thank you very much, mister. Merged.

PR #447, opened 2015-11-06
[WIP] Does this build?

gjvis:


gjvis:
`hub` opened this against the wrong upstream, apologies

PR #456, opened 2015-12-09
add link to behave-django and behave

nikolas:
These are alternatives to lettuce that are more actively maintained, which I use with splinter.


andrewsmedina:
@nikolas thank you!

PR #461, opened 2016-01-20
Issue 460 define __ne__ for Python 2.7.x compatibility

bharath23:
According to Python 2.7.x data model [1]:
"There are no implied relationships among the comparison operators. The
truth of x==y does not imply that x!=y is false. Accordingly, when
defining **eq**(), one should also define **ne**() so that the operators
will behave as expected."

[1] https://docs.python.org/2/reference/datamodel.html

### DIFF
--- a/splinter/request_handler/status_code.py
+++ b/splinter/request_handler/status_code.py
@@ -16,6 +16,9 @@ class StatusCode(object):
     def __eq__(self, other):
         return self.code == other
 
+    def __ne__(self, other):
+        return self.code != other
+
     def __str__(self):
         return "{} - {}".format(self.code, self.reason)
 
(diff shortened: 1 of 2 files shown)


andrewsmedina:
Hi @bharath23 thank you for contribute. Can you add a test for this PR? 

If you need help, let me know.

bharath23:
Hi @andrewsmedina what would be the best way to run a single test. When I use the run_tests.py, I need to run the entire test suite for testing if my test works. That will help me fix this the test suite.

andrewsmedina:
@bharath23 you can run tests for a specific driver. My suggestion is to run the `zope` driver tests:

```
./run-tests.py -w tests/test_zopetestbrowser.py
```

bharath23:
@andrewsmedina  I was trying to modify tests/status_code.py to test this. Seems like ZopeTestBrowser doesnt have the status_code object. I guess I will use firefox tests.

bharath23:
@andrewsmedina I think I worked out the kinks in testing.

kolapsys:
Hi @andrewsmedina, I think this ticket can be closed too due to PR #645 merge. :)

PR #463, opened 2016-02-06
Support for is_element_present_by_* in non-javascript drivers.

dryobates:
In order to bring more compatibility between javascript and non-javascript drivers I have implemented missing is_element_present_by_* methods. Although those methods are supposed to be used for dynamically appearing elements on page they are convenient also for static elements. I thing it is worth including them for non-javascript drivers. We have already is_text_present in non-javascript drivers.

Implemented methods will raise NotImplementedError if someone try to use wait_time, as it has not sense to use it in non-javascript drivers. I'm not really sure is it expected behavior. My thinking was that it is better to clearly communicate for user that it's not supported and make him catch error and do what he thing is suitable then force him to debug why it doesn't wait for loading.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -96,6 +96,7 @@ specific docs for each driver):
 * :doc:`Phantomjs WebDriver </drivers/phantomjs>`
 * :doc:`zope.testbrowser </drivers/zope.testbrowser>`
 * :doc:`django client </drivers/django>`
+* :doc:`flask client </drivers/flask>`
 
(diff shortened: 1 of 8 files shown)


andrewsmedina:
@dryobates I believe that ignore wait_time is better than raises a `NotImplementedError`.

dryobates:
OK, changed.

andrewsmedina:
@dryobates thank you!

PR #464, opened 2016-02-08
Consistent redirects behaviour

dryobates:
I have moved redirection test to base as it should work exactly the same in all drivers.
For flask, redirection have to be done in driver because flask doesn't expose redirection chain yet. It can be dropped (and probably integrated with django driver) once flask will expose it.


andrewsmedina:
@dryobates thank you!

PR #479, opened 2016-05-20
Webdriver support for clearing text field types

infame-io:
Clears text field types if they already have values in them

`browser.find_by_name('telephone').clear()`
`browser.find_by_tag('textarea').clear()`
`browser.find_by_id('username').clear()`


douglascamata:
Can you add some tests, please?

douglascamata:
Looks good. Thanks for the contribution!

andrewsmedina:
thank you for contribute. but, the documentation is missing and before to add a method in a driver that method should be added in the base driver api https://github.com/cobrateam/splinter/blob/master/splinter/driver/__init__.py#L14

PR #481, opened 2016-06-02
Add executable_path to phantomjs driver

bmcculley:
Enable passing the executable path as an argument for the phantomjs driver.


douglascamata:
@bmcculley thanks for the contribution. Can you add docs about this feature, please?

bmcculley:
@douglascamata definitely, I'll get on that later tonight and submit the pull request. Thank you for the feed back.

douglascamata:
@bmcculley no need to submit another PR, just update this one and it's ok. 👍

bmcculley:
Alright, I updated the docs. Let me know if I should fix anything else.

docs/drivers/phantomjs.rst
```diff
@@ -35,6 +35,15 @@ the ``Browser`` instance:
     from splinter import Browser
     browser = Browser('phantomjs')
 
+Phantomjs can also be used from a custom path. To do this you need to pass the string ``phantomjs`` with the ``executable_path`` when creating the ``Browser`` instance:
```
douglascamata: @bmcculley I suggest a clearer description here: PhantomJS can also be used from a custom path. To do this you give the path to the PhantomJS executable using the keyword argument ``executable_path`` when creating the ``Browser`` instance.

bmcculley:
@douglascamata excellent, it's much better like that. Thank you :+1:

tests/test_webdriver_phantomjs.py
```diff
@@ -69,6 +70,70 @@ def test_can_work_on_popups(self):
         # we can implement this test
         pass
 
+class PhantomJSExecutablePathTest(WebDriverTests, unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        lib_path = os.path.abspath(os.path.join(os.path.dirname( __file__ ), '..', 'lib'))
```
douglascamata: @bmcculley aren't you forgetting to commit this `lib` folder? I don't see it anywhere.

bmcculley:
@douglascamata I wasn't sure about committing it, it contains the phantomjs executable which would be different depending on the system. So, maybe just some documentation explaining what needs to be in there? or idk?

douglascamata:
@bmcculley I guess it will always be there. I'm just worried about the builds on Travis. Need to fix them first, they're all hanging on Google Chrome's installation because of some package version conflict.

douglascamata:
@bmcculley I triggered another build of this pull request. I'll just wait for the results to see if your test passes and then we are good to merge. Sorry for the build trouble. 👍

bmcculley:
@douglascamata I don't think it's going to pass. It hangs on all the phantomjs tests, I might be wrong but I don't see any place where it gets installed. I think that needs to be addressed first.

douglascamata:
@bmcculley I see. Feel free to add PhantomJS installation to the build. 

Remember that you need to rebase your fork of splinter from the master branch of the original repository to have the TravisCI fixes, otherwise the build will still fail for this PR, as it already did when I manually triggered it again.

douglascamata:
@bmcculley update: no need to install phantomjs, it's already there according to https://docs.travis-ci.com/user/gui-and-headless-browsers/#Using-PhantomJS.

bmcculley:
@douglascamata awesome, thank you for the link. I'll look into this a little more.

bmcculley:
@douglascamata it's something with the tests themselves, it starts running but gets hung up and eventually travis-ci cancels the test because of it

```
$ ./run_tests.py -w ${DRIVER}

............EEE.

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

The build has been terminated
```

douglascamata:
@bmcculley can you reproduce on your machine? I'll try to reproduce on mine.

bmcculley:
@douglascamata Finally! It's because travis-ci has phantomjs 1.9.7. The test ran fine on my machine with 2.1.1 so I adjusted the `.travis.yml` to download the 2.x version and update the path before running the tests. I'll update this branch momentarily.

bmcculley:
@douglascamata a quick update, it's working now. But, I ended up just moving the phantomjs download to my github because bitbucket blocked the download. I'm not sure what would be the best way to go about it; but that should probably be moved to somewhere else in the future.

douglascamata:
@bmcculley I think we can host the it in Github itself using [Git Large File Storage](https://git-lfs.github.com). I'll investigate.

bmcculley:
@douglascamata nice :+1: that looks like it would a good option.

douglascamata:
@bmcculley please, can you remove the extracted phantomjs from the lib folder and add the compressed versions from http://phantomjs.org/download.html using git-lfs? Then in the build script use git-lfs again to fetch those files and extract them. This way we can merge this quickly.

bmcculley:
Definitely, I think it should be all set for this now.

I've also started a `dev` branch which has a lot more changes. Just need to finish writing a few new tests for the chrome driver and I can submit a pull request for that one.

douglascamata:
Great! Thank you.

douglascamata:
@bmcculley i'm sorry for taking so long to review this. I had an emergency flight to PyData Paris. Will review it tomorrow (I swear).

splinter/driver/webdriver/phantomjs.py
```diff
@@ -35,7 +35,13 @@ def __init__(self, user_agent=None, load_images=True,
         if desired_capabilities:
             capabilities.update(desired_capabilities)
 
-        self.driver = PhantomJS(desired_capabilities=capabilities, **kwargs)
+        if executable_path is not None:
```
andrewsmedina: instead to add `if` and `else` you cand add `executable_path` to kwargs parameters, something like this:

```
if executable_path:
    kwargs["executable_path"] = executable_path
```

btw this already works without this code, because splinter already gets custom parameters with **kwargs and pass it to PhantomJS.

bmcculley:
@andrewsmedina excellent :+1: , although it's not clear that it works like that (I had no idea).

So, would it be better to just update the documentation to show that splinter can be used like this and just drop this "added" functionality?

bmcculley:
@douglascamata no problem, hope it went well! :+1:

andrewsmedina:
I believe that only update the documentation is the best way.

bmcculley:
I going to go ahead and close this since it's not useful.

PR #483, opened 2016-06-03
google chrome install

bmcculley:
Gets Google chrome to install on travis-ci.org


douglascamata:
@bmcculley great! Thank you again. I see on the builds that there are some stuff that we need to fix on tests, but it's very good to have builds going again.

PR #488, opened 2016-06-23
update custom driver docs

bmcculley:
Update the documentation so it's obvious of how to use custom executable paths.


douglascamata:
Thank you one more time for the contributions!

PR #489, opened 2016-06-23
add note that firefox is used by default

bmcculley:
I added a note to the index and tutorial page to let users know that firefox is the default driver.


andrewsmedina:
@bmcculley thank you!

Issue #382, opened 2015-03-24
Chrome Preferences not passed to Selenium Driver


bgsmalls:
After trying to disable images using splinter, I discovered that the options are not properly passed to the Selenium Driver. Using the Selenium Chrome Driver works without an issue.

Specifically the experemental options are treated in an non-standard way by chrome.
Fortunatelly the issue is easy to fix.

/usr/local/lib/python2.7/dist-packages/splinter/driver/webdriver/chrome.py

OLD-with-Bug, line 17-20:

```
def __init__(self, user_agent=None, wait_time=2, fullscreen=False,
                 **kwargs):

        options = Options()
```

NEW-Bugfix, line 17-20:

```
def __init__(self, user_agent=None, wait_time=2, fullscreen=False, options=None,
                 **kwargs):

        options = options or Options()
```

Could you please add this fix to the main development ? Thanks a lot.

coanor:
We really need such a option to passed to `splinter.Browser()`,  one of mine is `--disable-setuid-sandbox`, see stack overflow [thread](http://stackoverflow.com/questions/37696854/splinter-how-to-add-chrome-options).

bmcculley:
I have introduced the passing of options as a list in my [dev branch](https://github.com/bmcculley/splinter/tree/dev) if you would like to have a look.

For example:

 ```
from splinter import Browser

options = ['--start-maximized', '--disable-setuid-sandbox']
browser = Browser('chrome', chrome_options=options)
```

bmcculley:
This is addressed in pull request #486

PR #494, opened 2016-07-16
allow to pass a chrome Options instance to Browser

nlaurance:
My use case was to enable the device emulation mode in Chrome.
Not much of a technical change, some more doc.


andrewsmedina:
@nlaurance thank you!

nlaurance:
just realized it should solve #382

PR #499, opened 2016-08-06
Don't hold a screenshot file opened

gentoo90:
Currently every taken screenshot file stays opened until python process is closed, which prevents removing it (on Windows)


andrewsmedina:
@gentoo90 thank you!

PR #503, opened 2016-08-10
Enable timeout setting for Firefox webdriver

rlafuente:
Currently there's no straightforward way to specify the timeout for the Firefox webdriver. 

Selenium's default value (30) is too short for low-performance settings, so it would be desirable to not only have a larger timeout window (this PR sets it to 90) as well as letting the user define this when instantiating the webdriver.


rlafuente:
Hey, just noticed this is still open. I tried looking through the Travis build logs to understand why it failed but I can't make much sense of the logs. Can someone point me the way to fix whatever problem this PR caused?

andrewsmedina:
@rlafuente thank you!

Issue #125, opened 2012-01-05
ValueError using zope.testbrowser


oinopion:
I'm trying to use zope.testbrowser, but I keep getting this error:

```
Traceback (most recent call last):
  File "/home/oinopion/.virtualenvs/meetupon/lib/python2.6/site-packages/lettuce/core.py", line 117, in __call__
    ret = self.function(self.step, *args, **kw)
  File "/home/oinopion/Work/meetupon/meetupon/accounts/features/user-steps.py", line 24, in i_press
    button = world.browser.find_by_value(button_value).first
  File "/home/oinopion/.virtualenvs/meetupon/lib/python2.6/site-packages/splinter/driver/zopetestbrowser.py", line 116, in find_by_value
    return self.find_by_xpath('//*[@value="%s"]' % value, original_find="value", original_selector=value)
  File "/home/oinopion/.virtualenvs/meetupon/lib/python2.6/site-packages/splinter/driver/zopetestbrowser.py", line 103, in find_by_xpath
    return self.find_by_name(xpath_element.name)
  File "/home/oinopion/.virtualenvs/meetupon/lib/python2.6/site-packages/splinter/driver/zopetestbrowser.py", line 127, in find_by_name
    control = self._browser.getControl(name=name, index=index)
  File "/home/oinopion/.virtualenvs/meetupon/lib/python2.6/site-packages/zope/testbrowser/browser.py", line 383, in getControl
    label, name, self.mech_browser.forms(), include_subcontrols=True)
  File "/home/oinopion/.virtualenvs/meetupon/lib/python2.6/site-packages/zope/testbrowser/browser.py", line 389, in _get_all_controls
    onlyOne([label, name], '"label" and "name"')
  File "/home/oinopion/.virtualenvs/meetupon/lib/python2.6/site-packages/zope/testbrowser/browser.py", line 85, in onlyOne
    "Supply one and only one of %s as arguments" % description)
ValueError: Supply one and only one of "label" and "name" as arguments
```

It is caused by this line:

```
world.browser.find_by_value(button_value).first
```

It works correctly with firefox driver.

konstantinkobs:
This still doesn't work. I call `browser.find_by_id('logIn_btn')` and get the same error message as @oinopion. It seems like if zope.testbrowser thinks that special characters are spaces...

matclab:
`find_by_value` calls find_by_xpath which itself does:
```
    for xpath_element in html.xpath(xpath):
[...]

PR #508, opened 2016-08-22
zopebrowser: don't find_by_name if there ain't no name

w0de:
Ref: #125 

ZopeTestBrowser.find_by_xpath now checks if xpath_element has a name before passing it to ZopeTestBrowser.find_by_name.

Later: can probably maybe add logic to check for labels, since zope.testbrowser.browser.getControl accepts a name or a label.


andrewsmedina:
hi @w0de, thanks for contribute. 

Can you add a test for this PR?

Issue #509, opened 2016-08-26
status_code is only ever 200 - OK.


lexual:
See attached testcase.

A status_code of 200 seems to be hard-coded since this commit?

https://github.com/cobrateam/splinter/commit/d151c565c1068c6f0ac3f09ff010353b9b1a9de2

```
import splinter

with splinter.Browser('firefox') as browser:
    # should be 200
    browser.visit('http://google.com')
    print(browser.status_code.code)
    print(browser.status_code.is_success())

    browser.visit('http://google.com/is_this_a_bug')
    print(browser.status_code.code)
    print(browser.status_code.is_success())

    browser.visit('http://github.com/is_this_a_bug/is_this_a_bug_project')
    print(browser.status_code.code)
    print(browser.status_code.is_success())

    browser.visit('http://this_domain_doesnt_even_exist.coma')
    print(browser.status_code.code)
    print(browser.status_code.is_success())
```

Gives this output:

```
$ python ./testcase.py 
200
True
200
True
200
True
200
True
```

This should be returning 404 for all but the 1st example.

bmcculley:
Not sure if it's an acceptable fix, but I just created a PR #510 which uses httplib to address this bug. It makes another request for the HEAD of the page and returns the status code and reason.

The example test case above now returns:

```
$ python testcase.py
301
True
404
False
301
True
301
True
```

bmcculley:
Sorry, I didn't really look at the old removed code before doing this. That was a much better implementation but pretty much the same thing.

What about running splinter with a proxy server by default that would deliver the status code without needing to make another request?

andrewsmedina:
How this proxy server will work?

bmcculley:
So, I think it would require quite a bit of rewriting splinter to get it to work, firefox I believe needs to use the hub server in order to get it to work with a proxy. Chrome and phantomjs can be set fairly easily, not sure about the others though.

I was thinking it would be possible to run the proxy as a subprocess and pass the http status code back to splinter. I'm thinking it could use openssl to handle secure connections. I'm not sure if something like this would be acceptable to add to the project since it would require so much refactoring, more overhead, and extra components to make it all work.

samuelcolvin:
Do we have an update on when this is going to be fixed? 

For me this is about the most serious issue imaginable with splinter: the primary test of server response is incorrect such that tests are likely to pass when they should fail! 

:ghost: :scream: :imp: :hankey: :bug: :japanese_ogre: :no_good:

[...]

PR #510, opened 2016-08-27
update for issue #509

bmcculley:
An update to address the status_code bug, uses httplib to request the HEAD of the requested page and get the http status code and reason.


samuelcolvin:
I don't think this works because the server may respond to `HEAD` differently to how it would respond to the subsequent `GET` or `POST` etc. request.

Issue #511, opened 2016-08-27
Chrome options


bmcculley:
Seems like the way the chrome drive sets options should be updated to avoid interfering with other possible arguments being passed to the function. Unless I'm missing something.

PR #512, opened 2016-08-27
update the way chrome options are passed (issue #511)

bmcculley:
Allows the [various chrome options](http://peter.sh/experiments/chromium-command-line-switches/) to be passed in as a list

```
from splinter import Browser

options = ['--lang=en-us', '--start-fullscreen']

with Browser('chrome', options=options) as browser:
    browser.visit('http://example.com')
    browser.quit()
```

### DIFF
--- a/splinter/driver/webdriver/chrome.py
+++ b/splinter/driver/webdriver/chrome.py
@@ -17,18 +17,22 @@ class WebDriver(BaseWebDriver):
     def __init__(self, options=None, user_agent=None, wait_time=2, fullscreen=False, incognito=False,
                  **kwargs):
 
-        options = Options() if options is None else options
+        pass_options = Options() 
+
+        if options is not None:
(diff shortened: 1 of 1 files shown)


andrewsmedina:
@bmcculley why the list is better than an `Option` ?

bmcculley:
@andrewsmedina I'm not sure that my thinking is correct, but my thought was, the only way to build the options to pass is:

```
options = Options()
options.add_argument("some_arg")
```

or possibly looping through a list:

```

option_list = ['option_1', 'option_2']
options = Options()
for opt in option_list:
    options.add_argument(opt)
```
so doing this way cuts down on the amount of code that is needed each time by putting the code to build the options inside the driver.

Anyway, not sure if this is a good idea or if I'm missing something?

andrewsmedina:

Issue #513, opened 2016-08-30
chrome default value for incognito


nicolasunravel:
signature

``    def __init__(self, options=None, user_agent=None, wait_time=2, fullscreen=False, incognito=False,
                 **kwargs):
``

doesn't match following test ``if incognito is not None:``

so default is to activate incognito mode

andrewsmedina:
fixed by @https://github.com/cobrateam/splinter/pull/514

PR #514, opened 2016-08-30
#513 remove incognito mode by default

oleksiiunravel:
fix bug #513


andrewsmedina:
@oleksiiunravel thank you!

PR #522, opened 2016-10-22
consider a <textarea> a control element

afeld:
Previously, `find_by_xpath()` (or `find_by_tag()`, etc.) would not consider a `<textarea>` a "control element", and would return it as an `LxmlElement`, which can't be `fill()`d. This (should) make it so that all `find_by_*` methods treat `<textarea>`s the same.

### DIFF
--- a/splinter/driver/lxmldriver.py
+++ b/splinter/driver/lxmldriver.py
@@ -266,7 +266,7 @@ class LxmlDriver(ElementPresentMixIn, DriverAPI):
         return element.tag == 'a'
 
     def _element_is_control(self, element):
-        return hasattr(element, 'type')
+        return hasattr(element, 'type') or element.tag == 'textarea'
 
     @property
     def cookies(self):
--- a/splinter/driver/lxmldriver.py
+++ b/splinter/driver/lxmldriver.py
@@ -266,7 +266,7 @@ class LxmlDriver(ElementPresentMixIn, DriverAPI):
         return element.tag == 'a'
 
     def _element_is_control(self, element):
-        return hasattr(element, 'type') or element.tag in ['button', 'input', 'textarea']
+        return element.tag in ['button', 'input', 'textarea']
 
     @property
     def cookies(self):
--- a/splinter/driver/lxmldriver.py
+++ b/splinter/driver/lxmldriver.py
@@ -266,7 +266,7 @@ class LxmlDriver(ElementPresentMixIn, DriverAPI):
         return element.tag == 'a'
 
     def _element_is_control(self, element):
-        return hasattr(element, 'type') or element.tag == 'textarea'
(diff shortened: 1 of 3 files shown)


splinter/driver/lxmldriver.py
```diff
@@ -266,7 +266,7 @@ def _element_is_link(self, element):
         return element.tag == 'a'
 
     def _element_is_control(self, element):
-        return hasattr(element, 'type')
+        return hasattr(element, 'type') or element.tag == 'textarea'
```
afeld: Seems like the former should also be checking for `element.tag == 'input'`, but I didn't want to mess with it.

tests/form_elements.py
```diff
@@ -31,6 +31,11 @@ def test_can_choose_a_radio_button(self):
         self.browser.choose("gender", "M")
         self.assertTrue(self.browser.find_by_id("gender-m").checked)
 
+    def test_can_find_textarea_by_tag(self):
+        "should provide a way to find a textarea by tag_name"
+        tag = self.browser.find_by_tag("textarea").first
+        self.assertEqual(None, tag.value)
```
afeld: Wasn't sure of the best way to test this change...open to ideas!

afeld:
Heh, just ran into the same problem right after this, except with a `<button>`. Fixed!

splinter/driver/lxmldriver.py
```diff
@@ -266,7 +266,7 @@ def _element_is_link(self, element):
         return element.tag == 'a'
 
     def _element_is_control(self, element):
-        return hasattr(element, 'type')
+        return hasattr(element, 'type') or element.tag == 'button' or element.tag == 'textarea'
```
andrewsmedina: I believe that:

```
element.tag in ['button', 'textarea'] is a better option.

File: splinter/driver/lxmldriver.py [hunk 4]
```diff
@@ -266,7 +266,7 @@ def _element_is_link(self, element):
         return element.tag == 'a'
 
     def _element_is_control(self, element):
-        return hasattr(element, 'type')
+        return hasattr(element, 'type') or element.tag in ['button', 'input', 'textarea']
```
afeld: Given that this is now accepting all `<input>`s, I'm not sure there's a need for checking if it has a `type` attribute.

andrewsmedina: I believe that the type check is unecessary

andrewsmedina:
@afeld thank you!

afeld:
Hey there! Just wondering if you were planning to cut a new release containing this fix anytime soon?

andrewsmedina:
I believe that we can release the 0.7.5 with these changes. I will try to do it in this week.

andrewsmedina:
@afeld released in the splinter 0.7.5

PR #527, opened 2016-11-30
update README project links (pytest-splinter, PyPOM and pypom_form)

davidemoro:
Add new project links:
* pytest-splinter
* PyPOM
* pypom_form


andrewsmedina:
@davidemoro thank you

PR #529, opened 2016-12-14
initial suggestion for select_by_text instead of using value

### LINKED ISSUE #528: fill_form for selects should use option text and not value
Values of options should not be known by users of a browser, but the options text should be.

https://github.com/cobrateam/splinter/blob/1c08cbe59830d8004a5bf397391c586c78b9fa74/splinter/driver/webdriver/__init__.py#L435

peterlauri:
Closed #528

### DIFF
--- a/splinter/driver/webdriver/__init__.py
+++ b/splinter/driver/webdriver/__init__.py
@@ -432,7 +432,7 @@ class BaseWebDriver(DriverAPI):
                     if field.value == value:
                         field.click()
             elif element._element.tag_name == 'select':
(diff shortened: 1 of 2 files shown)


andrewsmedina:
@peterlauri thank you

peterlauri:
But I didn't write any tests yet :)

andrewsmedina:
@peterlauri this code is already covered by `fill_form` tests :)

peterlauri:
Sweet :)

PR #542, opened 2017-05-12
Document behavior of `slowly=True` in `type()`

neilsh:
I was surprised to find that passing `slowly=True` changes the function's behavior so that it doesn't perform the action directly, but instead returns an iterator. Hopefully this will help keep others from being surprised.


douglascamata:
@neilsh thanks for the contribution. 👍

PR #548, opened 2017-07-03
Headless option for chrome

danclaudiupop:
Starting with Chrome 59, Google has implemented a headless option. 
As you all know, PhantomJS will no longer be supported as the project lead stepped down.

More infos here:
https://developers.google.com/web/updates/2017/04/headless-chrome

Regards,
Dan


douglascamata:
@danclaudiupop do you mind updating docs as well?

andrewsmedina:
LGTM. But, in the future we can think a better way to handle these browser specific parameters.

andrewsmedina:
@danclaudiupop thank you!

PR #553, opened 2017-07-29
Include a submit button value when form submitted by clicking 

femdom:
I discovered that lxml driver doesn't include a value when a form is submitted by clicking a submit button.

I fixed it for Django and chrome clients. Other tests don't run on my machine for now. 

Please, let me know what do you think.


andrewsmedina:
@femdom thank you

PR #556, opened 2017-08-04
handle exceptions when calling quit on webdriver

### LINKED ISSUE #555: unnecessary exception thrown when quitting gecko driver
Using gecko driver 0.18 on an ubuntu installation with firefox 53.0 and when running the simple splinter test example provided in the docs, the following exception is thrown. 

`E       selenium.common.exceptions.WebDriverException: Message: quit`

It seems that when the quit function is called on the driver, gecko driver is throwing an exception instead of exiting gracefully. 

Please see this stack overflow link for a similar description of the issue

https://stackoverflow.com/questions/45505441/webdriverexception-message-quit

lesfleurs:
In reference to issue 555
Fixes #555

### DIFF
--- a/splinter/driver/webdriver/__init__.py
+++ b/splinter/driver/webdriver/__init__.py
@@ -13,6 +13,7 @@ from contextlib import contextmanager
 
 from selenium.common.exceptions import NoSuchElementException
 from selenium.webdriver.common.action_chains import ActionChains
+from selenium.common.exceptions import WebDriverException
 
 from splinter.driver import DriverAPI, ElementAPI
 from splinter.element_list import ElementList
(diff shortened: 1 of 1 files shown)


splinter/driver/webdriver/__init__.py
```diff
@@ -473,7 +474,10 @@ def select_by_text(self, name, text):
         self.find_by_xpath('//select[@name="%s"]/option[text()="%s"]' % (name, text)).first._element.click()
 
     def quit(self):
-        self.driver.quit()
+	try:
```
andrewsmedina: can you fix the indentation here?

lesfleurs: Of course. Updated.

andrewsmedina:
@lesfleurs thank you

PR #562, opened 2017-09-04
Add example remote URL for Saucelabs

Zowie:
I ran into not knowing exactly what the URL should be. After checking the Saucelabs docs I found out, might be handy to just have it in here.


andrewsmedina:
@Zowie thank you!

PR #563, opened 2017-09-12
Chrome driver download source seems to be invalid

mpibpc-mroose:
The given URL to the chrome driver for windows seemed to be orphan. Added the "official" URL to the chrome driver, which I found to be working.


andrewsmedina:
@mpibpc-mroose thanks for contribute

PR #585, opened 2018-02-10
Allow return value for execute_script to be returned. Fixes #584.

### LINKED ISSUE #584: webdriver execute_script discards result
The [current implementation of execute_script](https://github.com/cobrateam/splinter/blob/c2f65f47ad0ff3a630f3bd4dbc1d0000a99f4412/splinter/driver/webdriver/__init__.py#L197) discards the result from `driver.execute_script`. Instead, it should pass through the result.

jaraco:
As described in #584, webdriver.execute_script should relay any return value from the underlying `driver.execute_script` call.

### DIFF
--- a/splinter/driver/webdriver/__init__.py
+++ b/splinter/driver/webdriver/__init__.py
@@ -195,7 +195,7 @@ class BaseWebDriver(DriverAPI):
         self.driver.refresh()
 
     def execute_script(self, script):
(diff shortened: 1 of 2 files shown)


andrewsmedina:
I believe that you shoud add a test to cover it. If you need help let me know :)

andrewsmedina:
@jaraco  thanks for contribute

PR #587, opened 2018-02-17
docs text fix

bclindner:
missed a line break to make the code block display properly.


andrewsmedina:
@bclindner thanks for contribute

PR #590, opened 2018-03-03
Pass 'options' instead of 'chrome_options'

StabbarN:
'options' are replaced by 'chrome_options' in Selenium's Chrome webdriver.

See https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/chrome/webdriver.py#L50-L51
```
if chrome_options:
            warnings.warn('use options instead of chrome_options', DeprecationWarning)
            options = chrome_options
```


andrewsmedina:
@StabbarN thank you

Vieites:
Thank you.

Vieites:
@StabbarN Thank you

PR #601, opened 2018-04-10
Windows Variables Suggestions

DonRichards:
Referencing Windows XP instructions might alarm some people and give the impression they have found a legacy repo. Although the URL is not an official Windows page I believe it holds the information needed. I would suggest moving the information to the wiki in the future.


andrewsmedina:
@DonRichards thanks for contribute :)

PR #607, opened 2018-04-27
Fix typo in driver error message ("executiong" => "execution")

jdanford:


andrewsmedina:
@jdanford thanks for contributing

Issue #595, opened 2018-03-22
splinter 0.7.7 breaks 0.7.5 functionality


sureshvv:
It seems buttons HAVE to be inside forms in 0.7.7.

Otherwise I get:

```
Traceback (most recent call last):
   File "/home/suresh/appt/myproject/appt/tests/test_blackout2.py", line 96, in test_delete
    browser.find_by_name('login').click()
  File "/home/suresh/appt/local/lib/python3.5/site-packages/splinter/driver/lxmldriver.py", line 397, in click
    return self.parent.submit_data(parent_form)
  File "/home/suresh/appt/local/lib/python3.5/site-packages/splinter/driver/djangoclient.py", line 121, in submit_data
    return super(DjangoClient, self).submit(form).content
  File "/home/suresh/appt/local/lib/python3.5/site-packages/splinter/driver/lxmldriver.py", line 74, in submit
    data = self.serialize(form)
  File "/home/suresh/appt/local/lib/python3.5/site-packages/splinter/driver/lxmldriver.py", line 48, in serialize
    form.remove(input)
  File "src/lxml/lxml.etree.pyx", line 950, in lxml.etree._Element.remove (src/lxml/lxml.etree.c:50799)
ValueError: Element is not a child of this node.
````

andrewsmedina:
it looks like a bug. I will try to reproduce it

sureshvv:
The html:

[bug.zip](https://github.com/cobrateam/splinter/files/1975367/bug.zip)

The code:

          from splinter import Browser
          browser = Browser('django')
          browser.visit('/login')
          browser.find_by_id('user-email').fill('jacob')
          browser.fill('password', 'top_secret')
          browser.find_by_name('login').click()

PR #611, opened 2018-05-08
fix issue #595

sureshvv:
have created issue before. See https://github.com/cobrateam/splinter/issues/595


sureshvv:
The test failures seem to be because the tests could not find "geckodriver" on the path. Is this a problem with the travis set up?

andrewsmedina:
@sureshvv thanks for contribute

sureshvv:
Happy that contribution is accepted.


________________________________

Issue #605, opened 2018-04-17
Support Django 2.0 test client as backend


Incanus3:
Hi,
we've recently migrated to Django 2.0. Now I wanted to create a high-level test using splinter with django backend, but found out, that I can't even install splinter[django], since it depends on django<1.12. Is there something I can do to support Django 2.0?

andrewsmedina:
We should add the support for Django 2.0

andrewsmedina:
closed by #615

PR #615, opened 2018-06-04
Django 2

theunraveler:
After some pretty basic testing, it looks to me like splinter supports Django 2 without any modification, so I just removed the upper constraint.

Related to https://github.com/cobrateam/splinter/issues/605.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ env:
   - DJANGO_VERSION=1.9 DRIVER=tests/test_djangoclient.py
   - DJANGO_VERSION=1.10 DRIVER=tests/test_djangoclient.py
   - DJANGO_VERSION=1.11 DRIVER=tests/test_djangoclient.py
+  - DJANGO_VERSION=2.0.6 DRIVER=tests/test_djangoclient.py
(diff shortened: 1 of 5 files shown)


theunraveler:
Test failures seem unrelated...

andrewsmedina:
I believe that just make possible to install django 2 is not enough. It is necessary to update django in test-requirements and https://github.com/cobrateam/splinter/blob/master/test-requirements.txt and add a env in travis to test it: https://github.com/cobrateam/splinter/blob/master/.travis.yml#L28

theunraveler:
OK, updated the requirements and build matrix. It looks like all of the relevant tests are passing (particularly `DJANGO_VERISON=2.0.6` with `PYTHON=3.4`). Since Django 2 does not support Python 2, I excluded that from the matrix.

There are quite a few other failing tests, but they appear to be unrelated to the changes in this PR.

theunraveler:
Anything else I can do to push this PR forward? Thanks!

theunraveler:
Nevermind, looks like it might have been picked up in https://github.com/cobrateam/splinter/commit/907208296d1936b2e8abfe21b7109cb0dad9e6a2#diff-2eeaed663bd0d25b7e608891384b7298. Thanks!

PR #619, opened 2018-07-04
Update chrome.rst

mdaz78:
Change the command for installation of chromedriver using brew.


andrewsmedina:
@mdaz78 thanks for contribute

PR #623, opened 2018-07-28
fix docstring typo in splinter.driver

scresante:
one letter typo in splinter/driver/__init__.py


andrewsmedina:
@scresante thanks for contributing

PR #629, opened 2018-08-28
Improve Chrome WebDriver on Linux docs

nlhkabu:


andrewsmedina:
@nlhkabu thanks for contribute!

PR #634, opened 2018-09-25
firefox: fix DeprecationWarning with firefox_options

blueyed:
Fixes:

> DeprecationWarning: use options instead of firefox_options


andrewsmedina:
@blueyed thanks for contribute

PR #640, opened 2018-10-04
Fix travis config for python 3.7

paulohrpinheiro:
In this [travis report](https://github.com/travis-ci/travis-ci/issues/9815), is a fix to the error with python 3.7. This PR apply this fix in splinter.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,11 @@ cache: pip
 language: python
 python:
   - "2.7"
-  - "3.7"
+matrix:
+  include:
+    - python: 3.7
+      dist: xenial
+      sudo: true
 env:
   - DJANGO_VERSION=1.7.11 DRIVER=tests/test_djangoclient.py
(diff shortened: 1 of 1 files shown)


andrewsmedina:
Why to use matrix is better than just add the 3.7 in the Python versions list?

paulohrpinheiro:
Hello. My intention was to test the given solution, to change the image used in the test and a parameter, without causing errors in the tests for Python 2.7. If there is no objection, the configuration is much clearer as it was, changing the image used for all tests.

paulohrpinheiro:
Hello. My intention was to make the slightest possible change by trying to affect only the versions that were in error. Since the solution involves changing the image used, I did it that way. But if there is no objection in changing this image, this change can be made globally, without having to create an entry for version 3.7 only.

andrewsmedina:
Hi @paulohrpinheiro thanks for this PR. :)

Issue #460, opened 2016-01-19
status_code requires __ne__ to work in python 2.7.x


bharath23:
Quoting to the 2.7.x data model [1]

"There are no implied relationships among the comparison operators. The truth of x==y does not imply that x!=y is false. Accordingly, when defining **eq**(), one should also define **ne**() so that the operators will behave as expected."

So in python 2.7.x one cannot do the following

browser = Browser()
browser.visit("http://www.google.com")
if browser.status_code != 200:
    print("ERROR")

This however will work in python 3.x because the data model says **ne** will automatically invert the return code of **eq**

[1] https://docs.python.org/2/reference/datamodel.html

andrewsmedina:
closed by #645

PR #645, opened 2018-10-07
Add __ne__ to the StatusCode class for Python 2.7.x compatibility

kolapsys:
Provide solution for issue #460


andrewsmedina:
@kolapsys thanks for contribute

kolapsys:
I think this also solves issue #461

PR #646, opened 2018-10-08
Update django and travis tests

bmcculley:
This pull request adds django 2 for python 3 and adjusts the tests to accommodate.

Updated test-requirements.txt so that it can handle both python 2 and 3
which eliminated the need for the py3 requirements. Also updated the
setup.py file to use environment markers to handle the different django
versions depending on python version.

Moved the loading of chrome and firefox browsers to the travis.yml file
as addons. Also, moved geckodriver into the travis.yml file, being loaded
from the bash scripts seemed to be introducing errors.

The test setup was changed to handle the different django versions, I
removed the DJANGO_VERSION environment variable from the tests where this
wasn't necessary. The ones where it was were moved under their respective
python versions.


andrewsmedina:
@bmcculley thank you

PR #647, opened 2018-10-08
Update switch to alert

bmcculley:
The syntax for switch to alert has changed from
`switch_to.alert()` to `switch_to_alert()`, updated webdriver init to reflect this change.


andrewsmedina:
@bmcculley thank you!

PR #650, opened 2018-10-09
from fake_webapp import read_static

Allhailankurgupta:
imported read_static from fake_webapp in tests/base.py


andrewsmedina:
@Allhailankurgupta why this is necessary?

PR #652, opened 2018-10-10
Update run_tests.py

bittukumarray:
Added a comment regarding __name__=='__main__'


andrewsmedina:
@bittukumarray whas is the objectibe of this PR. Why this comment is really necessary?

Issue #639, opened 2018-10-04
Cleaning up errors in documentation


boltzmann-brain:
Skimming the docs, I notice a lot of language mistakes, typos, and outdated information. Does anyone mind if I work on cleaning those up? I'm a professional technical writer and I've worked with Sphinx.

jsfehler:
A lot of these issues have been fixed. If anything new is found it can be addressed in a new issue.

PR #653, opened 2018-10-13
Minor text correction, slight changes to readability

Murkantilism:
Made with regard to #639


andrewsmedina:
@Murkantilism thank you

Issue #608, opened 2018-04-28
Finding partial text


Qqwy:
It is very common that you want to look for a snippet of text, rather than a text element that as a whole contains exactly what you look for; like when looking for a username on a page.

The current `browser.is_text_present` however only matches a text element 1:1. (and the documentation about this is not very clear that this is what is going on).

So if we have

```
<h1>Welcome, DeathSlayer!</h1>
```

Then `browser.is_text_present("DeathSlayer")` will return false, as will `browser.is_text_present("Welcome, DeathSlayer")` (because of the missing exclamation mark).

Could another command be introduced for this?

BhavishyaJain:
@andrewsmedina Hi, is anyone working on this? if not, then can I take it up?

Vichoko:
Bump this. I think is a crucial feature, because text is now always known completely. For achiving this, maybe regex could be a very nice input to represent lookup in a more flexible way.

abathur:
I was having trouble matching some text and saw this issue while looking for possible explanations. It may just be a version difference (I'm using 0.11.0) since this was opened, or perhaps I'm using a different driver, but I don't see this behavior. 

I can match partial text just fine. 

FWIW, the problem I was having was with matching text that spanned more than one element. I looked at the implementation and see that it's just calling roughly `splinter.driver.find_element_by_tag_name("body").text.index(needle)`. In my case, checking the contents of `splinter.driver.find_element_by_tag_name("body").text` helped me identify what I needed to match against.

LexaDS:
Do not know if this issue was solved, but what about the contains() function in XPath, might help with the whole partial text issue.

sridhar-mani:
is this solved?

PR #655, opened 2018-10-13
Took a crack at #608

Murkantilism:
I'm on a windows machine so I cannot set up a dev environment to test this myself, if @Qqwy or someone else can test this implementation for me I would appreciate it.

### DIFF
--- a/splinter/driver/lxmldriver.py
+++ b/splinter/driver/lxmldriver.py
@@ -330,7 +330,7 @@ class LxmlDriver(ElementPresentMixIn, DriverAPI):
         try:
             body = self.find_by_tag("body").first
             # Check if there's a substring match at all
-            if body.find(text) is not -1:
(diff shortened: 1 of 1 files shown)


splinter/driver/lxmldriver.py
```diff
@@ -317,7 +317,29 @@ def is_text_not_present(self, text, wait_time=None):
...
+        while time.time() < end_time:
+            if self._contains_text(text):
+                return True
+        return False
+    
+    def _contains_text(self, text):
+        try:
+            body = self.find_by_tag("body").first
+            # Check if there's a substring match at all
+            if body.find(text) is not -1:
```
42B: Why are you checking for identity here instead of equality?

```python
if body.find(text) != -1:
```

andrewsmedina:

PR #656, opened 2018-10-14
Made full TOC appear in sidebar for better navigation. Various language edits.

qwermish:
Before, the full TOC was not available in the sidebar, so if you clicked on a link to the Installation page (for example), you would have to navigate back to Home to get to any other page. I updated  doc-requirements.txt to use the latest Read the Docs theme, which allows for captions in the TOC. Then, I put a few toctree directives in index.rst so that the full TOC would appear in the sidebar, with captions separating the different categories.

I also made various language edits for clarity, grammar, and punctuation.

Finally, Sphinx was throwing some warnings about title underlines being too short. I don't think they affected the build, but they were extra noise within legitimate warnings, so I made those underlines longer to get rid of the warnings.

### DIFF
--- a/doc-requirements.txt
+++ b/doc-requirements.txt
@@ -1,2 +1,2 @@
-Sphinx==1.2.3
-sphinx-rtd-theme==0.1.6
+Sphinx>=1.7.8
+sphinx-rtd-theme>=0.1.8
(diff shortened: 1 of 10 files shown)


andrewsmedina:
@qwermish thank you!

Issue #628, opened 2018-08-27
Unable to chain find_by_css().find_by_text() with 0.9.0


jacebrowning:
https://github.com/cobrateam/splinter/issues/426 seems to have resurfaced in 0.9.0.

With 0.8.0, `browser.find_by_css(".onboarding-modal").find_by_text("Slack")` finds one element on my site, but 0.9.0 returns none.

I think https://github.com/cobrateam/splinter/commit/1cfd05086a5ccb1c3297072e2bc0e713e71b2156 is the commit that changed this behavior.

I can try to find a reproduceable example on a public site if necessary.

rubiagatra:
Can I help with this one @jacebrowning ?

rubiagatra:
Thanks! @jacebrowning I will install and follow the tutorial first, If I am facing something that I don't know I will contact you

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 0.05 ETH (11.23 USD @ $224.66/ETH)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/cobrateam/splinter/628/1376).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/cobrateam/splinter/628/1376).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $47,781.96 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 12 months from now.
Please review their action plans below:


**1) [Naggertooth](https://gitcoin.co/profile/Naggertooth) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/cobrateam/splinter/628/1376?mutate_worker_action=approve&worker=Naggertooth) | [reject worker](https://gitcoin.co/issue/cobrateam/splinter/628/1376?mutate_worker_action=reject&worker=Naggertooth))_.**

Hi
I have time to work on this
I'll read the changelog and fix it.

[...]

Issue #612, opened 2018-05-13
element.find_by_text searches whole dom


jaraco:
When trying to do a contextual search for text, the result returned includes all matching text for the whole DOM instead of just those nodes that are children of the contextual node.

jaraco:
I believe the issue lies here https://github.com/cobrateam/splinter/blob/master/splinter/driver/webdriver/__init__.py#L407.

The `//*` means all nodes in dom. It should be `.//*` to mean all child nodes.

andrewsmedina:
@jaraco this line is the `browser.find_by_text` and in this case the code is right. I believe that the bug is here in the `element.find_by_text`: https://github.com/cobrateam/splinter/blob/ed21fb5495965d265bd5bd3e95b0dfb7121da7c2/splinter/driver/webdriver/__init__.py#L618

jaraco:
Yes. That makes sense.

PR #658, opened 2018-10-18
fix find_by_text chaining

super-oleg:
this seems like a fix for #628


andrewsmedina:
We should check if this fix will not come back this bug: https://github.com/cobrateam/splinter/issues/612

andrewsmedina:
@super-oleg thank you!

PR #665, opened 2018-11-08
Add support for retrieval of response headers to Flask client

sampart:
This is functionality I've needed when using Splinter with behave to test web apps.

### DIFF
--- a/splinter/driver/__init__.py
+++ b/splinter/driver/__init__.py
@@ -596,6 +596,15 @@ class DriverAPI(InheritedDocs("_DriverAPI", (object,), {})):
             "%s doesn't support cookies manipulation" % self.driver_name
         )
 
+    @property
+    def response_headers(self):
+        """
+        Headers from the most recent response.
+        """
+        raise NotImplementedError(
+            "%s doesn't support access to the response headers." % self.driver_name
+        )
+
 
 class ElementAPI(InheritedDocs("_ElementAPI", (object,), {})):
     """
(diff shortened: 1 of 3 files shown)


bernardofire:
Hi @sampart .
Do you think it's necessary to create a new method? Wouldn't call `_response.headers` solve the case?

sampart:
Hi @bernardofire.  You're right that this is possible at present by calling `_response.headers`.  However, `_response` is presumably an internal field as it starts with an underscore, so it would be better to have a public way of accessing this.  This leaves you free to change the underlying implementation in future without breaking backwards-compatibility.

bernardofire:
Hi @sampart ! Fair point, but can you give a bit more context on what was the exactly behavior you were trying to test? Splinter is meant to be an acceptance test tool, to simulate an user; which means that what you have on the screen should be enough to make the validations.

sampart:
Here's a bit of context - the sort of things I've used this for in other frameworks in the past:

* Testing APIs (where headers are often particularly significant)
* Retrieving the status code. This can be particularly useful to ensure that the 404 page isn't misidentified as a simple real page with similar content, and for testing redirects, but it's also helpful to explicitly state that success is expected in certain cases
* Testing security concerns - e.g. that security headers are correctly set on certain routes

However, if this change doesn't fit with your direction for the project then please do feel free to close the PR - I won't mind!

bernardobarreto:
closing this PR for now

PR #679, opened 2019-04-08
Fix broken use of switch_to.alert

jsfehler:
Selenium's switch_to.alert is an attribute, not a method.


jsfehler:
#684 fixes this issue.

PR #682, opened 2019-06-11
Browser.get_alert() returns Alert instead of a wrapper object

jsfehler:
- Remove `fill_with` method, since it just wraps `send_keys`

The purpose of this PR is to change Browser.get_alert() such that it provides access to the real selenium Alert object instead of a wrapper. The context manager functionality has been preserved.


andrewsmedina:
I believe that we should keep the `fill_with` as deprecated. And remove it in a futher version.

jsfehler:
@andrewsmedina Agreed. I've rebased the branch with this change.

andrewsmedina:

PR #683, opened 2019-06-12
ElementList is no longer a subclass of list

### LINKED ISSUE #680: ElementAPI.clear() shadows list.clear() in python 3.3+
The following code in python <= 3.2 will call ElementAPI.clear():

```
browser.find_by_css('foobar').clear()
```

However, in python 3.3 and forward, the built-in list implements a clear() method. ElementList is a subclass of list, and as such the above code will call ElementList.clear().

The following still calls ElementAPI.clear() in all python versions:

```
browser.find_by_css('foobar').first.clear()
```

ElementList probably should not subclass list directly.

jsfehler:
Fixes #680

### DIFF
--- a/splinter/element_list.py
+++ b/splinter/element_list.py
@@ -7,7 +7,7 @@
 from splinter.exceptions import ElementDoesNotExist
 
 
-class ElementList(list):
+class ElementList(object):
     """
     List of elements. Each member of the list is (usually) an instance
     of :class:`ElementAPI <splinter.driver.ElementAPI>`.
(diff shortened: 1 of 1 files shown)


jsfehler:
@andrewsmedina The changes in this PR fix the majority of the remaining test failures. If you don't have any concerns, merging will help to cut down the noise when looking at the remaining failures.

andrewsmedina:
@jsfehler I'm testing this in my local machine as part of review process. Finishing this step I will approve and merge this pr.

andrewsmedina:

PR #684, opened 2019-06-12
Browser.get_alert() now waits for alert to present

jsfehler:
Fixes failing alert tests


andrewsmedina:

PR #685, opened 2019-06-13
Fix test_clicking_submit_input_posts_empty_value_if_value_not_present

jsfehler:
In Chrome, the default value for an `input type="submit` is "Submit". However in Firefox its "Submit Query".


andrewsmedina:

PR #686, opened 2019-06-13
Do not implicitly rely on selector name

### LINKED ISSUE #681: WebDriverElement. select* implicitly rely on presence of name attribute for selector
#620 asked about using `select` (and by extension, `select_by_text`) for an element without a name

But if you check implementation for methods defined for `WebDriverElement`, both contain search by selector's `name`

https://github.com/cobrateam/splinter/blob/32f11ff7fd7841b123f157720dcc95740f156ca8/splinter/driver/webdriver/__init__.py#L616

https://github.com/cobrateam/splinter/blob/32f11ff7fd7841b123f157720dcc95740f156ca8/splinter/driver/webdriver/__init__.py#L621

I want to make a pull request for fixing this behavior, but I'm not sure which would be better:
- completely remove `name` from xpath 
- use xpath with `name` if `self["name"]` is not empty, otherwise, use xpath without `name`

aviskase:
Fixes #681

### DIFF
--- a/splinter/driver/webdriver/__init__.py
+++ b/splinter/driver/webdriver/__init__.py
@@ -622,12 +622,12 @@ class WebDriverElement(ElementAPI):
 
     def select(self, value):
         self.find_by_xpath(
-            '//select[@name="%s"]/option[@value="%s"]' % (self["name"], value)
+            '//select/option[@value="%s"]' % value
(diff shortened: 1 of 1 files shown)


andrewsmedina:
@aviskase thanks for contribute

PR #687, opened 2019-06-13
Fix tests failures due to django2 compatibilty

aviskase:
This should enable to normally run tests for django2

Always red tests are bad)


andrewsmedina:
@aviskase thanks for contribute

PR #688, opened 2019-06-14
Fix test_capabilities_set. 

jsfehler:
acceptSslCerts has been deprecated in favour of acceptInsecureCerts. However, the current implementation of webdriver will always override a desired capability in favour of an Options value, and the default Options object has this value set to a true value.


andrewsmedina:

PR #689, opened 2019-06-14
Fix test_double_click

jsfehler:
This change prevents the page from scrolling to the top after the first click.


andrewsmedina:

PR #690, opened 2019-06-19
Fix mouse_out method

jsfehler:
Changes the behaviour of the mouse_out() method to move to (-10, -10) relative to the top left of the element.

+ Add a new test where the hovered element is at the top-left of the browser window.

This fixes the mouse_out behaviour in Chrome.


andrewsmedina:

PR #691, opened 2019-06-19
Fix test_webdriver_chrome test_attach_file

jsfehler:
Convert the bytes object into a string for python 3.6+ compatibility.


andrewsmedina:

PR #692, opened 2019-06-19
Fix popups tests

jsfehler:
Some tests fail in Chrome due to a leftover popup window from a previous test.

The reason appears to be that if a popup window is already open, each browser has different behaviour when opening a new popup after a URL redirection.
- Chrome: A new window is opened.
- Firefox: The existing popup window is used.

This can be reproduced manually using the example app:

Steps to Reproduce:
1 - Run the example app and go to localhost:5000
2 - Open a popup window
3 - Navigate to google.com
4 - Navigate to localhost:5000 (Do not use a back button)
5 - Open a popup window

### DIFF
--- a/tests/popups.py
+++ b/tests/popups.py
@@ -38,6 +38,9 @@ class PopupWindowsTest(object):
         self.assertTrue(self.browser.windows.current.is_current)
         self.assertFalse(self.browser.windows.current.next.is_current)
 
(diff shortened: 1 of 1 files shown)


andrewsmedina:
@jsfehler thanks for contribute. Now the tests are green! :clap: :clap: :clap:

PR #696, opened 2019-06-25
ElementList now accepts slice objects

### LINKED ISSUE #519: Can't slice result of find_by_tag
Here's example straight from my code:

```
all_raw_msgs = self.browser.find_by_id('messages').first.find_by_tag('li')
new_raw_msgs = all_raw_msgs[self.message_count:-1]
```

I get error 'slice(0, -1, None) is not JSON serializable', so I looked up source code and found this:

```
    def __getitem__(self, index):
        if not isinstance(index, int):
            return self.first[index]
        try:
            return super(ElementList, self).__getitem__(index)
        except IndexError:
            raise ElementDoesNotExist(
                u'no elements could be found with {0} "{1}"'.format(
                    self.find_by, self.query))
```

Why case of non-integer index is special? This is the line that makes attempt to slice fail. What is reason for this?
Thanks for your time.

jsfehler:
Fixes #519

### DIFF
--- a/splinter/element_list.py
+++ b/splinter/element_list.py
@@ -36,7 +36,7 @@ class ElementList(object):
         self.query = query
 
     def __getitem__(self, index):
-        if not isinstance(index, int):
+        if not isinstance(index, int) and not isinstance(index, slice):
             return self.first[index]
         try:
             return self._container[index]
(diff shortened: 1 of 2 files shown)


andrewsmedina:
@jsfehler Thanks for contributing.

Issue #520, opened 2016-10-20
how to access element in iframe that has no tag


aswwindkk:
i'd like to try the way ----with browser.get_iframe("**") as iframe:
but the iframe doesn't has a tag of id or name even index,how could i access the element in this kind of frame
 i want to click the date 20160925,the source code is something like this:
![qq 20161020154544](https://cloud.githubusercontent.com/assets/22768303/19550593/c17564c2-96dc-11e6-8a01-115e22d24dcb.png)
thanks

andrewsmedina:
closed in 0.11.0 release

PR #697, opened 2019-06-26
Allow browser.get_iframe() to accept a web element

jsfehler:
This mimics the functionality available in Selenium and solves #520


andrewsmedina:

PR #698, opened 2019-06-27
Add ability to take HTML snapshots

jsfehler:


andrewsmedina:

Issue #575, opened 2018-01-04
Can not chain find_link_by syntax to find_by


j7an:
I would like to use the following format to find a link within an xpath:

```
field = context.browser.find_by_xpath(xpath_location).find_link_by_partial_text(text)
    field.click()
```

But this fails with the following error:
```
AttributeError: 'ElementList' object has no attribute 'find_link_by_partial_text'
```

I can use the following alternative:
```
field = context.browser.find_by_xpath(xpath_location).find_by_text(text)
    field.click()
```

andrewsmedina:
@j7an thanks for report

andrewsmedina:
closed by https://github.com/cobrateam/splinter/pull/699

PR #699, opened 2019-07-02
Change find_link_by API and add functionality to Elements

### LINKED ISSUE #419: find_link_* methods cannot be chained to Element/ElementList
The Splinter docs say, "Finding methods are chainable," but this is not true for `find_link_*` methods. They are missing from Splinter's element classes. Please add them.

jsfehler:
Fixes #419 and #575

This adds the `find_link_by` functionality to elements in the form of Element.links.find_by_

The Browser behaves the same way, ie: `Browser.links.find_by_href` however the original API is preserved. `Browser.find_link_by_href` still works.

If this PR is accepted, I'll follow it up with changing the documentation and adding a deprecation notice for the Browser.find_link_by methods.

If the API change is rejected, I'll add the older method names to the Element. Regardless, I think this pattern of using a parent object, even if hidden, still reduces code duplication.

### DIFF
--- a/splinter/driver/webdriver/__init__.py
+++ b/splinter/driver/webdriver/__init__.py
@@ -179,12 +179,56 @@ class Windows(object):
         )
 
 
+class FindLinks(object):
+    """Contains methods for finding links in a parent.
+
+    Accessed through the browser object or an element via the links attribute.
+
+    Example:
+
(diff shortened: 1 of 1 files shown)


andrewsmedina:
@jsfehler thank you

PR #701, opened 2019-07-17
Add test environments for py37 and py38

jsfehler:


andrewsmedina:

PR #703, opened 2019-07-19
Use pytest as the test runner

jsfehler:
This PR replaces the current test runner logic in run_tests.py with the pytest library. Portions of that file have been moved to tests/conftest.py.

pytest is compatible with unittest, allowing most of the tests to remain untouched. New tests can take advantage of pytest's simpler assertion syntax.

The chrome and firefox tests have been altered to standardize the getting of Browser objects, and take advantage of pytest's fixtures.

This change allows all the different ways of kicking off tests to run the same way, through tox. The distribution of the tests on travis-ci is unchanged.


andrewsmedina:
@jsfehler this is a very nice improvement. Thank you!

PR #704, opened 2019-07-22
Add deprecation warning to find_link_by_ methods. Document browser.li…

jsfehler:
…nks and element.links


andrewsmedina:
@jsfehler thank you!

PR #707, opened 2019-07-29
Fix ElementList.__getattr__() to no longer hide ElementNotFound

vdboor:
When elements aren't found on the page, calling methods like ElementList.click() didn't raise this error. Instead it would report "AttributeError: 'ElementList' object has no attribute 'click'" which confuses users to look at other places than the actual error.


andrewsmedina:
@vdboor thanks for contribute.

PR #714, opened 2019-09-21
Fix Firefox headless mode to handle custom firefox_binary option

kjkorhonen:
This commit fixes Firefox headless mode so that it respects the possible `firefox_binary` option.

Previously a custom Firefox path string or browser object passed in `kwargs['firefox_binary']` would not work in headless mode. Instead the option was silently overwritten and a default executable location was used.

### DIFF
--- a/splinter/driver/webdriver/firefox.py
+++ b/splinter/driver/webdriver/firefox.py
@@ -4,6 +4,7 @@
 # Use of this source code is governed by a BSD-style
 # license that can be found in the LICENSE file.
 import os
+import six
 
 from selenium.webdriver import DesiredCapabilities, Firefox
 from selenium.webdriver.firefox.firefox_profile import FirefoxProfile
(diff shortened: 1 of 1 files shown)


splinter/driver/webdriver/firefox.py
```diff
@@ -17,6 +17,11 @@
 from selenium.webdriver.firefox.firefox_binary import FirefoxBinary
 from selenium.webdriver.firefox.options import Options
 
+try:
+    basestring
+except NameError:  # Python 3.x
+    basestring = str
```
andrewsmedina: splinter uses `six` to helps with the python 2/3 compatibility. So you should use `six.string_types` (https://six.readthedocs.io/#six.string_types) instead to handle the `basestring` compatibility here.

splinter/driver/webdriver/firefox.py
```diff
@@ -63,7 +68,13 @@ def __init__(
 
         if headless:
             os.environ.update({"MOZ_HEADLESS": "1"})
-            binary = FirefoxBinary()
+            if 'firefox_binary' in kwargs:
+                if isinstance(kwargs['firefox_binary'], str):
```
andrewsmedina: You should check if the `firefox_binary` isinstance of six.string_types

andrewsmedina:
@kjkorhonen Thank you for contribute

PR #715, opened 2019-09-22
Fix Firefox driver to respect headless option in subsequent calls

kjkorhonen:
This commit clears the `MOZ_HEADLESS` environment variable explicitly when the `headless` option is not set. Selenium passes the environment variables to Firefox as they are, and the environment variable alone is enough to cause a headless launch.

Previously a single headless launch would cause all subsequent launches be headless regardless the option being set or not.


andrewsmedina:
@kjkorhonen thank you for contribute

PR #716, opened 2019-09-23
How to fill in complex text values

miohtama:
Added an example how to use JavaScript to enter complex text through Splinter


andrewsmedina:
@miohtama thank you for contribute!

Issue #387, opened 2015-04-06
browser.get_alert() does not return "None" when there is no alert


njbennett:
When I call `browser.get_alert()` on a page without an alert, instead of `None`, it returns `NoAlertPresentException: Message: No alert is present`-- contrary to the documentation.

Ie., 

```
from splinter import Browser

browser = Browser()
browser.visit('https://www.google.com')
browser.get_alert()
```

It'd be nice for this to work as documented, since then I'd have a simple way to check that there isn't an active alert on the page, although it's not a huge problem.

Is the problem with the code, or with the documentation? It looks like `.get_alert()` is using a deprecated Selenium method, though I'm not sure if that's the cause of the issue or not.

haoflynet:
Now I get `selenium.common.exceptions.TimeoutException: Message:`

kjkorhonen:
@andrewsmedina, I think this issue can be closed now?

PR #717, opened 2019-09-25
Browser.get_alert() returns None if no alert exists

kjkorhonen:
A fix for issue #387. Have `browser.get_alert()` return `None` if no alert exists, instead of Selenium exception.


andrewsmedina:
@kjkorhonen thank you for contribute.

Issue #702, opened 2019-07-18
HTTP Error 403: request disallowed by robots.txt


Constantin07:
Running in python venv (MacOS) `Python 2.7.16`

Installed packages in venv:
```
splinter==0.11.0
zope.testbrowser==5.3.3
lxml==4.3.4
cssselect==1.0.3
mechanize==0.4.2
```

```python
from splinter import Browser

jenkins_url = 'some value here'

browser = Browser('zope.testbrowser')
browser.visit(jenkins_url)
```

getting back:

```
Traceback (most recent call last):
  File "jenkins_test.py", line 9, in <module>
    browser.visit(jenkins_url)
  File "/Users/homedir/Documents/Jenkins_Test/venv/lib/python2.7/site-packages/splinter/driver/zopetestbrowser.py", line 81, in visit
    self._browser.open(url)
  File "/Users/homedir/Documents/Jenkins_Test/venv/lib/python2.7/site-packages/zope/testbrowser/browser.py", line 252, in open
    self._processRequest(url, make_request)
  File "/Users/homedir/Documents/Jenkins_Test/venv/lib/python2.7/site-packages/zope/testbrowser/browser.py", line 276, in _processRequest
    resp = make_request(reqargs)
  File "/Users/homedir/Documents/Jenkins_Test/venv/lib/python2.7/site-packages/zope/testbrowser/browser.py", line 250, in make_request
    return self.testapp.get(url, **args)
  File "/Users/homedir/Documents/Jenkins_Test/venv/lib/python2.7/site-packages/webtest/app.py", line 335, in get
    expect_errors=expect_errors)
  File "/Users/homedir/Documents/Jenkins_Test/venv/lib/python2.7/site-packages/zope/testbrowser/browser.py", line 92, in do_request
    self._assertAllowed(req.url)
  File "/Users/homedir/Documents/Jenkins_Test/venv/lib/python2.7/site-packages/zope/testbrowser/browser.py", line 89, in _assertAllowed
    raise RobotExclusionError(url, 403, msg, [], None)
zope.testbrowser.browser.RobotExclusionError: HTTP Error 403: request disallowed by robots.txt
```

What am I doing wrong ?

mattfeury:
looks like it was removed here: https://github.com/cobrateam/splinter/commit/a7cbbf9efd6f987779eaad90f8af2d9758a42b1e

no clue why. @andrewsmedina was the author so may be able to shed some light

andrewsmedina:
[...]

PR #719, opened 2019-09-28
[Bugfix & Enhancement] Screenshot

yaquawa:
Hey, thanks for the impressive works!!

I just found some issues, hopefully this can be merged 😆

## Solved

1. When using the [Emulation](https://sites.google.com/a/chromium.org/chromedriver/mobile-emulation), viewport won't get resized.
2. Can't get the correct full window size when element positioned using the `position: absolute;`.
3. Lazy-load images won't get loaded when capturing the full document.
4. The image scale won't change on higher pixel ratio device(window.devicePixelRatio > 1) when taking screenshot of an element. (I have reported the issue [here](https://bugs.chromium.org/p/chromedriver/issues/detail?id=3154))
5. Added `capture_viewport` & `capture_viewport_as_base64` methods

## Enhancement

1. Use the chrome-driver's [implementation](https://w3c.github.io/webdriver/#take-element-screenshot) to capture the element's screenshot.
2. Added `capture_screenshot_as_base64` method to both `BaseWebDriver` and `WebDriverElement`.
3. Added a parameter `waiting_time` for waiting images to be loaded.

## Breaking changes
Deprecated `screenshot` method, the new method name is `capture_screenshot`(signature changed).

I think we shouldn't force user to place the image to the temp directory. It's the user's responsibility to determine where the saved file should go 😀
(not yet fixed the test case and docs, if you are ok for this behavior I'll go to fix it.)

### DIFF
--- a/splinter/driver/webdriver/__init__.py
+++ b/splinter/driver/webdriver/__init__.py
@@ -599,20 +599,35 @@ class BaseWebDriver(DriverAPI):
     def uncheck(self, name):
         self.find_by_name(name).first.uncheck()
 
-    def screenshot(self, name="", suffix=".png", full=False):
-
-        name = name or ""
-
-        (fd, filename) = tempfile.mkstemp(prefix=name, suffix=suffix)
-        # don't hold the file
-        os.close(fd)
-
+    def screenshot(self, filename, full=False, waiting_time=0):
         if full:
             self.full_screen()
+            # trigger the `scroll` event to ensure lazy-load images can be rendered.
+            self.execute_script("window.dispatchEvent(new Event('scroll'))")
 
-        self.driver.get_screenshot_as_file(filename)
+        if waiting_time > 0:
+            time.sleep(waiting_time)
+
+        result = self.driver.get_screenshot_as_file(filename)
+        
         self.recover_screen()
-        return filename
+
+        return result
+
+    def screenshot_as_base64(self, full=False, waiting_time=0):
+        if full:
+            self.full_screen()
+            # trigger the `scroll` event to ensure lazy-load images can be rendered.
+            self.execute_script("window.dispatchEvent(new Event('scroll'))")
+
+        if waiting_time > 0:
+            time.sleep(waiting_time)
+
+        result = self.driver.get_screenshot_as_base64()
+
+        self.recover_screen()
+
+        return result
 
     def select(self, name, value):
         self.find_by_xpath(
@@ -632,16 +647,28 @@ class BaseWebDriver(DriverAPI):
 
     def full_screen(self):
         self.ori_window_size = self.driver.get_window_size()
-        width = self.driver.execute_script("return Math.max(document.body.scrollWidth, document.body.offsetWidth);")
-        height = self.driver.execute_script("return Math.max(document.body.scrollHeight, document.body.offsetHeight);")
-        self.driver.set_window_size(width, height)
+        width = self.driver.execute_script("return Math.max(document.body.scrollWidth, document.body.offsetWidth, document.documentElement.scrollWidth);")
+        height = self.driver.execute_script("return Math.max(document.body.scrollHeight, document.body.offsetHeight, document.documentElement.scrollHeight);")
+        self.set_viewport_size(width, height)
 
     def recover_screen(self):
         if self.ori_window_size:
             width = self.ori_window_size.get('width')
             height = self.ori_window_size.get('height')
-            self.driver.set_window_size(width, height)
+            self.set_viewport_size(width, height)
             self.ori_window_size = None
+
+    def set_viewport_size(self, width, height):
+        device_metrics = {
+            'width': width,
+            'height': height,
+            'deviceScaleFactor': self.execute_script('return window.devicePixelRatio'),
+            'mobile': self.execute_script("return typeof window.orientation !== 'undefined'")
+        }
+
+        self.driver.execute_cdp_cmd('Emulation.setDeviceMetricsOverride', device_metrics)
+        
+        return self
 
     def html_snapshot(self, name="", suffix=".html", encoding='utf-8'):
         """Write the current html to a file."""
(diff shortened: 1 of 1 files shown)


splinter/driver/webdriver/__init__.py
```diff
@@ -599,20 +599,35 @@ def check(self, name):
     def uncheck(self, name):
         self.find_by_name(name).first.uncheck()
 
-    def screenshot(self, name="", suffix=".png", full=False):
+    def screenshot(self, filename, full=False, waiting_time=0):
```
douglascamata: It's still possible to keep the same signature like this:

```python
def screenshot(self, name="", suffix=".png", full=False, waiting_time=0):
        if full:
            self.full_screen()
            # trigger the `scroll` event to ensure lazy-load images can be rendered.
            self.execute_script("window.dispatchEvent(new Event('scroll'))") 
        if waiting_time > 0:
            time.sleep(waiting_time)

        if not name:
             (fd, name) = tempfile.mkstemp(prefix=name, suffix=suffix)
            os.close(fd)
        result = self.driver.get_screenshot_as_file(name)

        self.recover_screen()

        return result
``` 

What do you think?

yaquawa: Thanks for the review!

The return value should be the filename otherwise user can't find it.
but, I still think this should be the user's responsibility to specify the path, so make the code easier to understand and predictable also more low coupling.

I mean it can be more readable if you write it something like this:

```py
browser.screenshot(get_temp_filename())

# helper
def get_temp_filename(prefix=name, suffix='.png'):
    (fd, name) = tempfile.mkstemp(prefix=name, suffix=suffix)
    os.close(fd)

    return name
```

yaquawa: or maybe we can create the .png file in temp directory and then move it to the directory where user specified?

douglascamata: I understand that. I'm just trying to think of something that wouldn't break backwards compatibility. In my idea you can still overwrite the file a call like the one you gave `browser.screenshot('~/latest.png')`  would work as you expect.

About the path: while the user should be responsible for a path, we can still provide a default temporary path and if the user really cares they can specify it. 

About what the function should return: why we should return the file path if the user is giving the file path to the function? Having a file descriptor, for example, makes it very easy to compare the screenshot with a test fixture. 

If we're breaking backwards compatibility, we could change from using `tempfile.mkstemp` to use `tempfile.NamedTemporaryFile` (https://docs.python.org/2/library/tempfile.html#tempfile.NamedTemporaryFile) and return a higher level file object.

yaquawa: > In my idea you can still overwrite the file a call like the one you gave browser.screenshot('~/latest.png') would work as you expect.

I don't think this would work, the file name will change every time when calling the `tempfile.mkstemp`.

> About what the function should return: why we should return the file path if the user is giving the file path to the function?

I mean if you are going to use the `tempfile.mkstemp` then we should return the file path, because the file name changes every time. In this commit it return `True` if succeed or `False` if not(which is the value selenium returns).

douglascamata: > I don't think this would work, the file name will change every time when calling the tempfile.mkstemp.

This function is only called whenever you don't give a `name` parameter to the function. If a name is not given, `tempfile.mkstemp` doesn't run. 

I see the problem then. And I think there is no solution on how to move forward without breaking backwards compatibility. The backwards compatibility is important though. Changing functions signature and return values needs to be done slowly. 

We should:

* Leave the current functions as they are today and add a deprecation warning, like done in #702. 
* Implement your proposed logic in new functions, with a different name.
* Release a new version of Splinter with a minor version increment with your added feature.
* When the deprecation has been there for long enough in the old functions, on the next major release (where breaking changes can be introduced) we drop them. 

Does this sound like a good plan for you, @yaquawa?

yaquawa: Thanks! I'm all for that 👍🏼👍🏼

But I think we can just create another branch for major version, for example `0.11.x`, then keep the current code there, so we can just merge this into the `master`.
And then output the warning on branch `0.11.x`. (which I think it's the common way for handling such a case)

douglascamata: Can you explain your plan better with real versions, I don't understand it well? For example, my plan is:

* Release 0.12.0 with your fix (will require that the functions are renamed to keep both behaviors), keep the old code and add a deprecation warning on it
* On the release 0.13.0, we just remove the old code.

This is exactly how phantomjs got removed. It started to output a deprecation warning in 0.8.0 and was removed in 0.9.0.

yaquawa: For example like [Laravel](https://github.com/laravel/framework), they have the major versions for each branch.
So like what they do, we can create a branch called `0.11.x`; on that branch, this commit won't be merged at all, instead we add some warning to the `screenshot` method on that branch. For this commit, just merge it into the `master` branch.

douglascamata: I'm sorry, but our model is not exactly the same as Laravel. Why not proceed as I proposed? I don't understand why you don't like my proposal... it's the way we used to work with breaking backwards compatibility and deprecations.

yaquawa: I'm all ok for your proposal 👌
I just thought that the newly added method have to be renamed again may take a little effort too. Anyway I'm going to respect the way how you manage this repo!

douglascamata: Thanks for the patience, @yaquawa. I'm sorry it was not clear until now. I'm eager to merge your contribution :)

yaquawa: Thanks! @douglascamata
I've putted the old method back, and added the new one as `capture_screenshot`(which match the CDP's method name).

douglascamata:
@yaquawa hey, I was having some holidays for the last 2 weeks and a half. How is it going here? Just ping me whenever you think the PR is ready 💪

yaquawa:
Hi @douglasmiranda, all the changes I've made are described on the top thread of this topic😀I think this PR is ready to go!

splinter/driver/webdriver/__init__.py
```diff
@@ -600,19 +614,93 @@ def uncheck(self, name):
         self.find_by_name(name).first.uncheck()
 
     def screenshot(self, name="", suffix=".png", full=False):
+        warnings.warn('Deprecated, use `self.capture_screenshot()` instead.')
```
douglascamata: What do you think about using the form ```warnings.warn('Deprecated, use `self.capture_screenshot()` instead.', FutureWarning)``` to be more explicit about the intent of the warning? It comes from https://docs.python.org/3/library/exceptions.html#FutureWarning.

The same applies for the other usages.

yaquawa: Sorry for the delayed reviews here too..!
I'm all for it 😀

Issue #515, opened 2016-09-20
Suggested updates to the `FlaskClient` driver 


anthonyjb:
Hi - first off thank you for this wonderful library. I'm using it in conjunction with **pytest-bdd** and **Flask** and for the most part it's been plain sailing, however I am currently using a custom `FlaskClient` that overrides the `_do_method` to resolve 2 issues/traits I came across.
### `302` and `303` behaviour

By default most browser (correctly or incorrectly) will use the `GET` method when redirecting based on a `302` response, and `303`s (I believe) specifically address this confusion by declaring that the `303` response from a `POST` request is to a `GET` resource.

At the moment using the `FlaskClient` all `30X` responses retain the current method, so if a `303` response is received from a `POST` request then the `_do_method` will make another request using the `post` not the intended `get` method.
### `args` and `form` data

**Flask** provides access to data transmitted via `POST` or `PUT` through `request.form`, alternatively data transmitted in the URL (e.g via `GET`) is available through `request.args` (there's also `request.values` which allows you to access data in either but that's not important here).

Using the `FlaskClient` when you submit a form with a `method` attribute of `GET` (e.g `<form method="GET" ...>`) the request will be made as using the `GET` method but the data will be sent as if `POST`ed and will end up in the `request.form` attribute. As you can imagine this then causes  issues because the view being called is looking for the submitted data in `request.args`.

The cause of the issue is that data is always sent in the `_do_method` using the `data` keyword argument when calling the `func_method`, but for a `GET` request (I believe) the `data` keyword should not sent as `None` and the data should be set against the URL as a query.
### Fixing these issues

As I mentioned at the start of the issue I'm currently using a patched driver based on the existing `FlaskClient` to resolve these issues/traits. I've included the code below in case others come across the same issues in the short-term and as a suggestion as to how they might be fixed. If you'd like me to contribute the changes as a pull request I'm more than happy to have a look at doing so over the weekend. 

> **Note:** My patch is using Python3.4+, hence `from urllib import parse` so would need to be made compatible with 2.7+ though I think that would be a relatively easy task.

``` python
from urllib import parse

from splinter.browser import _DRIVERS
from splinter.driver import flaskclient

__all__ = ['FlaskClient']


class FlaskClient(flaskclient.FlaskClient):
    """
    A patched `FlaskClient` driver that implements more standard `302`/`303`
    behaviour and that sets data for `GET` requests against the URL.
    """

    driver_name = 'flask'

[...]

PR #721, opened 2019-10-02
Fix FlaskClient redirects

lfdebrux:
This is another attempt at getting a fix for cobrateam/splinter#515 by @anthonyjb merged into splinter; I've added some tests but otherwise this is basically a copy of #517.

### DIFF
--- a/splinter/driver/flaskclient.py
+++ b/splinter/driver/flaskclient.py
@@ -6,6 +6,12 @@
 
 from __future__ import with_statement
 
+try:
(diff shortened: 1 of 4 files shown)


lfdebrux:
I don't really understand why the tests are failing; for me they are failing locally on master too.

lfdebrux:
Okay I finally see what I did wrong here, I did some bad HTML 😆 It should be fixed now

andrewsmedina:
@lfdebrux thanks for contribute

Issue #567, opened 2017-10-12
Suggest add in docs the possibility to use selenium special Keys


XuanShine:
I'm new on using github tools.
We should add in the docs that we can use selenium special Keys:

e.g. (return key):
from selenium.webdriver.common.keys import Keys
ElementAPI.type(Keys.RETURN)

or add a reference to selenium docs for special keys

andrewsmedina:
+1

caoliang12121:
yes,yes,yes!!

ynx0:
I would like to help update the docs. Can I start working on this?

andrewsmedina:
@ynx0 yes. Fell free to contribute with this issue.

tetsucceed:
Hello, this is my first time trying to contribute to Open Source. So if this issue is still open i think it's good chance to start with this.

andrewsmedina:
@tetsucceed this issue is still open. So, fell free to work on it. If you nedd help, let us know.

jsfehler:
Fixed by #722

PR #722, opened 2019-10-02
add in docs the possibility to use selenium #567

tetsucceed:
Could I have missed something?


andrewsmedina:
@tetsucceed thanks for contribute

Issue #712, opened 2019-09-11
is_element_visible missing from API docs?


tony:
[`is_element_visible`](https://github.com/cobrateam/splinter/blob/04a6da307f76d872ef1c227c47509fc983e5315c/splinter/driver/webdriver/__init__.py#L201) is a very helpful function, but haven't figured out why / what would be the clearest way to make it available in the docs (Should we make all `Window` available?)

andrewsmedina:
A good place for the docs about this method is the `matchers` page: https://github.com/cobrateam/splinter/blob/master/docs/matchers.rst

sukhbeersingh:
Hi , I would like to work on this issue.

andrewsmedina:
@sukhbeersingh fell free to work on it. If you need help let us know.

sukhbeersingh:
Is the intent to just include the said function in the matchers page? Or is it to make a new page (windows.rst) and incorporate this in there?

andrewsmedina:
closed by #723

PR #723, opened 2019-10-08
added docs for is_element_visible methods

sukhbeersingh:
#712 Adds documentation for the two cases ( `css`,`xpath`) of `is_element_visible` in [Matchers](https://splinter.readthedocs.io/en/latest/matchers.html) page. 

Please let me know if I missed something.


andrewsmedina:
@sukhbeersingh thanks for contribute.

PR #725, opened 2019-11-01
Retry WebElement.click if Exception is thrown

jsfehler:
Sometimes, an element may be covered by another, but only for a limited amount of time. This change adds a retry mechanism to a click attempt.


andrewsmedina:

PR #727, opened 2019-11-06
find_by_x methods in WebDriverElement now uses retry mechanism

jsfehler:


andrewsmedina:

PR #732, opened 2019-11-18
is_not_present/visible returns True immediately after not finding any…

jsfehler:
…thing

By exposing wait_time in find_by_css and find_by_xpath, these values are set to 0 for is_not_present/visible. This will prevent re-trying something that we want to confirm is not happening. The other find_by methods now expose wait_time as well, just for consistency.


andrewsmedina:

PR #430, opened 2015-08-21
Enable keep_alive for remote webdriver

bubenkoff:
Helps with the tests stability

[Review on Reviewable](https://reviewable.io/reviews/cobrateam/splinter/430)

PR #733, opened 2019-11-19
Enable keep_alive for remote webdriver

jsfehler:
Same as #430, but without merge conflicts.

### DIFF
--- a/splinter/driver/webdriver/remote.py
+++ b/splinter/driver/webdriver/remote.py
@@ -27,7 +27,7 @@ class WebDriver(BaseWebDriver):
         abilities = getattr(DesiredCapabilities, browsername, {})
         abilities.update(ability_args)
 
-        self.driver = Remote(url, abilities)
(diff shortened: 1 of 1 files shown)


splinter/driver/webdriver/remote.py
```diff
@@ -27,7 +27,7 @@ def __init__(self, url=DEFAULT_URL, browser="firefox", wait_time=2, **ability_ar
         abilities = getattr(DesiredCapabilities, browsername, {})
         abilities.update(ability_args)
 
-        self.driver = Remote(url, abilities)
+        self.driver = Remote(url, abilities, keep_alive=True)
```
andrewsmedina: I believe that a better way is to receive a parameter that defines if the keep_alive is enabled (or not)

jsfehler: Sure. I've opened #734 with more expansive changes, but will match all of Remote WebDriver's arguments.

jsfehler:
Replaced with #734

PR #734, opened 2019-11-20
Accept all valid arguments for Remote WebDriver

jsfehler:
This changes the Remote Browser from accepting desired capabilities in the keyword args to accepting anything and sending to Remote Webdriver as keyword args. As a result, desired capabilities are now sent as a dictionary instead of keyword args.

Documentation updated to reflect the new behaviour.

### DIFF
--- a/splinter/driver/webdriver/remote.py
+++ b/splinter/driver/webdriver/remote.py
@@ -33,12 +33,11 @@ class WebDriver(BaseWebDriver):
 
         # If no desired capabilities specified, add default ones
         caps = getattr(DesiredCapabilities, browser_name, {})
-        if not kwargs.get('desired_capabilities'):
-            kwargs['desired_capabilities'] = caps
-        else:
+        if kwargs.get('desired_capabilities'):
             # Combine user's desired capabilities with default
             caps.update(kwargs['desired_capabilities'])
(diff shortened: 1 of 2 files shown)


splinter/driver/webdriver/remote.py
```diff
@@ -19,15 +19,28 @@ class WebDriver(BaseWebDriver):
...
+            browser_name = "INTERNETEXPLORER"
+
+        # If no desired capabilities specified, add default ones
+        caps = getattr(DesiredCapabilities, browser_name, {})
+        if not kwargs.get('desired_capabilities'):
+            kwargs['desired_capabilities'] = caps
+        else:
+            # Combine user's desired capabilities with default
+            caps.update(kwargs['desired_capabilities'])
+            kwargs['desired_capabilities'] = caps
```
andrewsmedina: This code can be simplified:

```
        if kwargs.get('desired_capabilities'):
            # Combine user's desired capabilities with default
            caps.update(kwargs['desired_capabilities'])
        kwargs['desired_capabilities'] = caps
```

jsfehler: @andrewsmedina Good catch. Fixed.

andrewsmedina:

PR #737, opened 2019-11-27
Update version of Firefox and geckodriver used for tests

jsfehler:


andrewsmedina:

Issue #674, opened 2019-01-28
Remote WebDriverElement seems more limited than it could be


cuducos:
According to the implementation of the [remote web driver](https://github.com/cobrateam/splinter/blob/master/splinter/driver/webdriver/remote.py) it will always use the `splinter.driver.webdriver.remote.WebDriverElement` as it's element class. This class, for instance, limits the possibility of certain actions, such as [mouse over](https://github.com/cobrateam/splinter/blob/master/splinter/driver/webdriver/remote.py#L40-L44).

It turns out that it is actually possible to perform a mouse over using Chrome remote driver. I'm coding this solution in a current project, it merely replaces `splinter.driver.webdriver.remote.WebDriverElement` by `splinter.driver.webdriver.WebDriverElement` in the remote web driver class:

```python
from splinter.driver.webdriver import WebDriverElement
from splinter.driver.webdriver.remote import WebDriver


class RemoteWebDriver(WebDriver):
    """A clone of the default remote web driver class, but using the non-remote
    web driver element class."""

    def __init__(self, *args, **kwargs):
        super(RemoteWebDriver, self).__init__(*args, **kwargs)
        self.element_class = WebDriverElement  # overrides the element class

browser = RemoteWebDriver(browser='chrome', url=DRIVE_URL)
browser.find_by_css('#tooltip').mouse_over()
```

That said, my hunch is that **maybe** we might be too restrict with the limitations imposed by the remote `WebDriverElement` class. From this experiment it seems that at least remote Chrome can handle mouse overs… does that make sense?

If so, I suggest a roadmap to address this issue:

- [x] Test which of the limitations (`mouse_over`, `mouse_out`, `double_click`, `right_click` and `drag_and_drop`) are still present in remote Chrome driver
- [x] Test which of the limitations (`mouse_over`, `mouse_out`, `double_click`, `right_click`  and `drag_and_drop`) are still present in remote Firefox driver
- [x] Refactor `splinter.driver.webdriver.WebDriverElement` to allow users to perform actions supported in the tests from steps 1 and 2 (we might write two different classes, but details of the implementation are to be discussed latter IMHO)

~~We can edit~~ I just editted this table with the results of the tests:

| Action | Remote Chrome | Remote Firefox |
|---|:-:|:-:|
| `mouse_over` | ✅  | ✅ |
| `mouse_out` | 🚫 | 🚫 |
| `double_click` | ✅ | ✅ |
| `right_click` | ✅ | ✅ |
[...]

PR #738, opened 2019-12-02
Allow ActionChains when using Remote WebDriver

jsfehler:
Newer versions of Remote Firefox support ActionChains.

This also deals with most of #674.

However, there's still no specific test configuration for Remote Chrome.


andrewsmedina:
@jsfehler thank you for contribute.

PR #739, opened 2019-12-03
find_by with 0 second wait_time only checks once

jsfehler:
#732 Did not actually cause the is_element_not_x functions to return immediately after nothing is found. The 0 second wait time was overridden by the global wait_time.

This PR changes things in such a way that 0 seconds for wait_time will cause find_by to only search once, instead of polling.

The _find() function will run once if 0 seconds is the wait_time, else it will run in a while loop until timeout.


andrewsmedina:

PR #740, opened 2019-12-04
Add driver attribute to WebDriverElement

jsfehler:
This fixes an issue where mouse interaction fails on nested elements. The `parent` attribute of a nested WebDriverElement is another WebDriverElement, which has no driver attribute.

ie: 

```
x = browser.find_by_xpath('...')
y = x.find_by_xpath('...')
y.double_click()
```


jacebrowning:
This is blocking me from using `0.12.0` so a patch release including this fix would be great! 😄

PR #741, opened 2019-12-04
Refactor forward/back tests to run in isolation

jsfehler:
Baby steps towards running tests in parallel


andrewsmedina:
Nice catch!

PR #742, opened 2019-12-04
Patch Remote WebDriver to add retry attempts. Add retry to Browser init.

jsfehler:
Shamelessly pulled from the pytest-splinter project. These patches add some much needed stability to Remote WebDriver runs, especially when running tests in parallel.

Personally, these sorts of things make more sense as part of Splinter than as part of an external pytest plugin. As another benefit, they also make running Splinter's own tests in parallel more stable.

### DIFF
--- a/splinter/browser.py
+++ b/splinter/browser.py
@@ -6,6 +6,15 @@
 
 import sys
 
+try:
+    from httplib import HTTPException
+except ImportError:
+    from http.client import HTTPException
+
+from urllib3.exceptions import MaxRetryError
+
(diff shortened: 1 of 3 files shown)


andrewsmedina:

mpasternak:
Very cool patch, but there is a small problem with the exception handling...

I'm getting `UnboundLocalError: local variable 'e' referenced before assignment`, same as people here https://stackoverflow.com/questions/59506553/unboundlocalerror-from-splinter ... 

Gonna work on a fix.

mpasternak:
Silly me. Already fixed.

PR #744, opened 2019-12-11
WebDriverElement find_by_x methods now take wait_time argument

jsfehler:


andrewsmedina:

Issue #748, opened 2019-12-20
 local variable 'e' referenced before assignment in DriverNotFoundError


alphydan:
I am trying to run a simple:
```
from splinter import Browser
browser = Browser('chrome', **executable_path)
```
I have downloaded the latest chromedriver from https://chromedriver.chromium.org/,  I did `chmod +x` the driver, saved it to the working directory and defined the executable_path as:

`executable_path = {'executable_path': os.getcwd() + '/chromedriver'}`
 
However, I get the error:

--------------------------------------------------------------------------
```
UnboundLocalError                         Traceback (most recent call last)
<ipython-input-13-92a2d755a7ca> in <module>
      1 from splinter import Browser
----> 2 browser = Browser('chrome', **executable_path)

~/.local/share/virtualenvs/some-app-Aojajyp7/lib/python3.6/site-packages/splinter/browser.py in Browser(driver_name, retry_count, *args, **kwargs)
     88         raise DriverNotFoundError("No driver for %s" % driver_name)
     89 
---> 90     return get_driver(driver, *args, **kwargs)

~/.local/share/virtualenvs/some-app-Aojajyp7/lib/python3.6/site-packages/splinter/browser.py in get_driver(driver, retry_count, *args, **kwargs)
     66             pass
     67 
---> 68     raise e
     69 
     70 

UnboundLocalError: local variable 'e' referenced before assignment
```

jsfehler:
@andrewsmedina I have a fix for this, will open a PR tomorrow.

andrewsmedina:
closed by #749

nmfzone:
When the patch release?

jacebrowning:
@andrewsmedina could we please get a new release containing this fix? 

I'd like to be able to automatically install the necessary WebDriver using https://github.com/SergeyPirogov/webdriver_manager, but the `UnboundLocalError` causes Splinter to crash too early to detect which drivers are available.

PR #749, opened 2019-12-22
Fix: Incorrect error thrown when missing chrome/geckodriver

jsfehler:
Addresses the error message seen in: #748


andrewsmedina:

Issue #743, opened 2019-12-11
adding cookies issues


jadeiX:
Hello, 
I am using Django with LiveServerTestCase. I was trying to load my chrome driver with a cookie that preloads a user session cookie. 
for example:
` cookie = {'path': '/', 'value': '9gl4p7wlrtgkw3dfer1vi2jievkks8f1', 'name': 'sessionid'}`
however, the cookie doesn't load properly, and the user is not logged in when i run 
```
browser.cookies.add(cookie)
browser.reload()
```
When I inspected the code at [splinter/driver/webdriver/cookie_manager.py](https://github.com/cobrateam/splinter/blob/master/splinter/driver/webdriver/cookie_manager.py) I found that the way cookies are added is by 
```
for key, value in cookies.items():
            self.driver.add_cookie({"name": key, "value": value})
```
resulting 
```
{'name': 'path', 'value': '/'}
{'name': 'value', 'value': '9gl4p7wlrtgkw3dfer1vi2jievkks8f1'}
{'name': 'name', 'value': 'sessionid'}
```
which didnt really work.
I have revised the add function such that a cookie dict is added as a whole using Selenium `add_cookie` function. 
This is the new function:
```
def add(self, cookies):
    if isinstance(cookies, list):
        for cookie in cookies:
            self.driver.add_cookie(cookie)
            return
    self.driver.add_cookie(cookies)
```

Hope this helps

jsfehler:
Fixed by #895

Issue #388, opened 2015-04-15
CookieManager.add does not support path etc (only name/value)


blueyed:
`selenium/webdriver/remote/webdriver.py` allows for additional keys, like `path`, `domain`, `secure` and `expiry`:

```
def add_cookie(self, cookie_dict):
    """
    Adds a cookie to your current session.

    :Args:
     - cookie_dict: A dictionary object, with required keys - "name" and "value";
        optional keys - "path", "domain", "secure", "expiry"

    Usage:
        driver.add_cookie({'name' : 'foo', 'value' : 'bar'})
        driver.add_cookie({'name' : 'foo', 'value' : 'bar', 'path' : '/'})
        driver.add_cookie({'name' : 'foo', 'value' : 'bar', 'path' : '/', 'secure':True})

    """
    self.execute(Command.ADD_COOKIE, {'cookie': cookie_dict})
```

But this is not supported through splinter's CookieManager:

```
class CookieManager(CookieManagerAPI):

    def __init__(self, driver):
        self.driver = driver

    def add(self, cookies):
        if isinstance(cookies, list):
            for cookie in cookies:
                for key, value in cookie.items():
                    self.driver.add_cookie({'name': key, 'value': value})
                return
        for key, value in cookies.items():
            self.driver.add_cookie({'name': key, 'value': value})
```

The workaround appears to be calling `driver.add_cookie` directly.

(There appears to be another issue with this workaround though, because the cookie does not seem to "stick", when `visit()` has not been called before - using the cookie from Django's `client` for admin login; I'll have to debug this. Looks like #244)

PR #752, opened 2020-01-06
Standardize CookieManager and accept all valid cookie arguments

### LINKED ISSUE #307: refactor `CookieManager.add` method
the problem related here should be fixed https://github.com/cobrateam/splinter/pull/300/files#r11098327

jsfehler:
- Add CookieManager.delete_all() method
- Standardize the init phase for CookieManager subclasses
- Remove ability to send a list to CookieManager.add(). This was undocumented and having an argument that takes multiple types seems like a bad idea.
- Rewrite CookieManager.add() to accept arguments for cookie parameters instead of a dict with a key/value for name and value parameters.

The change to CookieManager.add() is incompatible with the current function signature, unfortunately.

Note that flask does not use the samesite parameter, and zope does not use httponly and samesite parameters.

Fixes #307, #388, and probably #743, but it needs to be tested.

### DIFF
--- a/docs/cookies.rst
+++ b/docs/cookies.rst
@@ -23,7 +23,7 @@ To add a cookie use the add method:
 
 ::
 
-    browser.cookies.add({'whatever': 'and ever'})
+    browser.cookies.add('cookie name', 'cookie value')
 
 Retrieve all cookies
 --------------------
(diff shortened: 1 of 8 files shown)


jsfehler:
Replaced with #767

PR #754, opened 2020-01-14
Added retry_count to get_driver

lrjball:
It looks like Browser is taking `retry_count` as an argument
which is never used. Fixed by adding it to get_driver.


andrewsmedina:
@lrjball thanks for contribute. Can you add a test to ensure that the retry parameter is used by get_driver method?

lrjball:
@andrewsmedina I've added a test now. There may be a better way of doing it that I haven't thought of, so I'm happy to make a change to the function if there is a better suggestion. Thanks

andrewsmedina (approved):

andrewsmedina:
@lrjball thanks for contribute

PR #762, opened 2020-02-24
Add FindLinks api to non-webdrivers

kageurufu:
I was surprised when the LxmlDriver didn't support FindLinks, so i've added it to all the drivers


jsfehler:
@kageurufu Thanks for the contribution

PR #770, opened 2020-04-02
Reduce duplicated code in WebDriver.is_text_present and is_text_not_p…

jsfehler:
…resent


splinter/driver/webdriver/__init__.py
```diff
@@ -442,45 +442,43 @@ def get_alert(self, wait_time=None):
...
+        StaleElementReferenceException is thrown if the page changes
+        quickly.
+        """
+        try:
+            self.driver.find_element_by_tag_name("body").text.index(text)
+            return True
+        except (NoSuchElementException, StaleElementReferenceException):
+            pass
+        except ValueError:
+            return False
```
andrewsmedina, line 461: How the default value is `False` there is no need to separate `ValueError` from the another errors
```suggestion
        except (NoSuchElementException, StaleElementReferenceException, ValueError):
            pass
```

PR #774, opened 2020-05-01
Update setup.py

### LINKED ISSUE #667: Django >= 2.0.6 now required for python > 3.3
[This commit](https://github.com/cobrateam/splinter/commit/907208296d1936b2e8abfe21b7109cb0dad9e6a2) ([this part in particular](https://github.com/cobrateam/splinter/commit/907208296d1936b2e8abfe21b7109cb0dad9e6a2#diff-2eeaed663bd0d25b7e608891384b7298R30)) pins the django version to >= 2.0.6 if you're installing on python 3.3. As far as I know, Django 1.11 is still supported on python 3, and this dependency declaration prevents me from using django 1.11 with splinter if I want to be on python 3.3.

I think the correct thing to do here is just to declare a dependency on `django >= 1.7.11` regardless of the python version, since this library supports all django versions above that, and has little to do with the python version.

jlgimeno:
Removes pinning Django to 2.0.6+ if installing on Python 3.3 or above.  Django 1.11 supports those versions.

Closes #667

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,7 @@ setup(
     install_requires=["selenium>=3.141.0", "six"],
     extras_require={
         "zope.testbrowser": ["zope.testbrowser>=5.2.4", "lxml>=4.2.4", "cssselect"],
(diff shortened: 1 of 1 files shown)


andrewsmedina:
@jlgimeno thanks for contribute!

PR #786, opened 2020-07-31
upgrade Pillow to the latest version to avoid some security problems

andrewsmedina:
there are some security reports in the oldest version: https://github.com/advisories/GHSA-43fq-w8qq-v88h

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -11,4 +11,4 @@ flake8==3.5.0
 six==1.11.0
 twine==1.11.0
 pytest==4.6.4
-Pillow==6.2.2
+Pillow==7.2.0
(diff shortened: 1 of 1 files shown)


test-requirements.txt
```diff
@@ -11,4 +11,4 @@ flake8==3.5.0
 six==1.11.0
 twine==1.11.0
 pytest==4.6.4
-Pillow==6.2.2
+Pillow==7.2.0
```
cclauss, line 14:
```suggestion
Pillow==6.2.2; python_version < '3.5'
Pillow==7.2.0; python_version >= '3.5'
```

cclauss:
Please resolve conflicts.

test-requirements.txt
jsfehler: I don't know why I forgot you can pin versions, but yes this is the most correct approach.

Issue #446, opened 2015-10-30
Using browser.screenshot(name = '/var/aa', suffer = '.jpg') filename turn out to 'aaxiHXq8.jpg'


cillyfly:
Using browser.screenshot(name = '/var/aa', suffer = '.jpg') filename turn out to 'aaxiHXq8.jpg' I don't know why? need some help

markuszoeller:
IIUC the string `xiHXq8` is a randomly generated string which gets appended to the filename `aa` you provide in `browser.screenshot(name, suffer)`. I assume this is done to avoid to replace an already existing file with the same name. Just a guess.

jsfehler:
@andrewsmedina browser.screenshot uses tempfile.mkstemp to get a filename with a random portion to the filename. Is this the desired behaviour? In the event of the provided filename already existing, should we write a new file (current behaviour, due to random string) or throw an exception? Either way, overwriting the existing screenshot is definitely something to avoid.

phaibin:
This is very confused. Can you add a switch to use my real filename or add random postfix?

jsfehler:
Fixed by #949

PR #791, opened 2020-08-17
Fix the generated screenshot filename when the absolute path to a file is used

luzfcb:
Problem: If a an absolute path to the screenshot file is defined, Splinter doesn't respect it and append a random file name to the final filename.

This problem seems to have gone unnoticed for the past 7 years.
This is the line where the error occurs:
https://github.com/cobrateam/splinter/blob/d28efabdf64c45a55fb349980f134f5a5edbd1c9/splinter/driver/webdriver/__init__.py#L628

This is the demonstration of the bug

```python
Python 3.8.4 (default, Jul 21 2020, 11:47:41) 
[GCC 7.5.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import tempfile
>>> name = "/tmp/test.png"
>>> suffix = ".png"
>>> (fd, filename) = tempfile.mkstemp(prefix=name, suffix=suffix)
>>> filename
'/tmp/test.png7hfkpela.png'
```

The expected filename is `/tmp/test.png` but was returned `/tmp/test.png7hfkpela.png`

This pull-request fix that problem.

### DIFF
--- a/splinter/driver/webdriver/__init__.py
+++ b/splinter/driver/webdriver/__init__.py
@@ -623,11 +623,12 @@ class BaseWebDriver(DriverAPI):
 
     def screenshot(self, name="", suffix=".png", full=False):
 
-        name = name or ""
+        filename = name or ""
 
-        (fd, filename) = tempfile.mkstemp(prefix=name, suffix=suffix)
-        # don't hold the file
-        os.close(fd)
+        if not filename:
+            (fd, filename) = tempfile.mkstemp(prefix=name, suffix=suffix)
+            # don't hold the file
+            os.close(fd)
 
         if full:
             ori_window_size = self.driver.get_window_size()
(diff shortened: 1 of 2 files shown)


jsfehler:
This is currently as designed. See the discussion at #446

luzfcb:
@jsfehler Per the documentation, it is a bug

image

danr:
As a workaround until this is merged it seems like you can do:

```python
browser.driver.get_screenshot_as_file('test.png') # see cobrateam/splinter/pull/791
```

This will save the screenshot to `test.png` without making a temporary filename.

jsfehler:
#949 is an implementation of compromise fix for #446, which covers what this PR was trying to do.

In order to get a filename that does not have a random name, set the parameter `unique_file` to False:

```
filename = browser.screenshot(name="myfilename", unique_file=False)
>>> filename == "myfilename.png"
```

PR #799, opened 2020-08-25
Remove sending a list of cookie dicts to CookieManager.add()

jsfehler:
Adding cookies with a list, eg:

```
my_list = [{...}, {...}]
CookieManager.add(my_list)
```

has always been undocumented.  Having an argument that takes multiple types seems like a bad idea. If a user really wants to add multiple cookies, it's more likely they're doing:

```
for c in cookies:
    CookieManager.add(c)
```

### DIFF
--- a/splinter/cookie_manager.py
+++ b/splinter/cookie_manager.py
@@ -25,9 +25,9 @@ class CookieManagerAPI(InheritedDocs("_CookieManagerAPI", (object,), {})):
     def __init__(self, driver):
         self.driver = driver
 
-    def add(self, cookies):
+    def add(self, cookie):
         """
-        Adds a cookie.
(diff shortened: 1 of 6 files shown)


andrewsmedina:
We should not make break changes without a deprecation warning. So @jsfehler what do you think in 0.15 support the both ways and add a deprecation warning (an example here: https://github.com/cobrateam/splinter/blob/215451825d1a120e442f69402b578944f27753e6/splinter/driver/webdriver/__init__.py#L474) and in the 0.16 we can remove the old way?

jsfehler:
@andrewsmedina Makes sense to me. See #801

andrewsmedina (approved):

Issue #477, opened 2016-03-20
Opening a link in a new tab


begueradj:
Howdy,

With Selenium, it is easy to open a link in a new tab. 
Is this even possible with Splinter?
I dived deeply into Splinter API documentation but I did not find anything related to tab features.

Thank you in advance for any hints,
Bill BEGUERADJ

WillyWinkel:
about two years ago and no answer at all?

andrewsmedina:
You can do it in the same way that it is done in selenium. Something like:

```
# from selenium.webdriver.common.keys import Keys
element.fill(Keys.CONTROL + Keys.RETURN
```

rominf:
@andrewsmedina, I don't like this approach. It's not portable: on OS X you have to use `Cmd` instead of `Ctrl`. Also, it's implementation dependent - your solution works only if the element was extracted from the browser, powered by Selenium.

I propose to add a method to `WebDriver`:
```
browser.windows.new('https://google.com')
```

andrewsmedina:
Good idea @rominf

szafran98:
Can I try doing this?

jsfehler:
I think `new_tab` makes more sense than `new`, which is too close to `__new__` but too different in behaviour for my liking.

@szafran98 Go for it. Just make sure to add unit tests.

szafran98:
@jsfehler Yes, actually i'm doing it, but i have a little problem with makefile and i can't run tests
`
Running all tests...
tox -- 'tests'
make: tox: No such file or directory
make: *** [test] Error 1
`
I've tried with absolute path and still doesn't work

szafran98:
@jsfehler Am I right that the topic of the issue is a possibility to open link in a new tab not a window? If yes then I almost done it and wrote tests, but I think there is a problem. I'm not using Firefox daily and I recognized that it always open link in a new window. Do you know any way to edit config immediately after run or need i to find workaround?

andrewsmedina:
@szafran98 and @jsfehler I believe that his method should belong to the `windows` scope. My proposal is to use`browser.windows.visit` or `browser.windows.new`.  What do you think about it?

szafran98:
@andrewsmedina I can move it, but as i see in the docs browser.windows -> """ A class representing all open browser windows """ so how the program will know in which window it should open a new tab?

szafran98:
@andrewsmedina when i'm puting this method to Windows class then django and flask client have access to this method.

[...]

PR #800, opened 2020-09-06
Link in new tab (#477)

szafran98:
Added new_tab method to BaseWebDriver (#477 ). It needs to execute some javascript code to change browser preferences, because at default firefox open all links even with `_blank` in a new window. Every other method which i found doesn't changing preferences if browser already running. Other possibility is to add `firefox_profile.set_preference("browser.link.open_newwindow", 3)` in `firefox.py`.


szafran98:
I've run tests on fresh clone and there are still errors. How to exclude django and flask from tests, because i don't think that it is possible to evaluate javascript code in their client, but maybe i put the code in bad place.
```
FAILED tests/test_webdriver_firefox.py::FirefoxBrowserTest::test_clicking_submit_input_posts_empty_value_if_value_not_present - AssertionError: None is not true : 'submit-input-no-value: Wyślij ...
FAILED tests/test_webdriver_firefox.py::FirefoxBrowserFullScreenTest::test_clicking_submit_input_posts_empty_value_if_value_not_present - AssertionError: None is not true : 'submit-input-no-valu...
```

andrewsmedina:
@szafran98 how this method only makes sense for webdriver drivers, this test should be defined in `WebDriverTests ` (https://github.com/cobrateam/splinter/blob/8fff713ec2741dc9b98babe2a34bdc9666e96196/tests/base.py#L167) class  instead in the `BaseBrowserTests`.

szafran98:
@andrewsmedina hey, can you review this?

szafran98:
@andrewsmedina hey, can you give me any answer?

szafran98:
@jsfehler hey, have you contact with @andrewsmedina ? Can someone review it?

andrewsmedina:
@szafran98 sorry for the delay. I believe that this PR looks good to me. Can you sync your branch with the master?

.idea/workspace.xml
```diff
@@ -0,0 +1,106 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jsfehler, line 1: Please remove this file from your PR

jsfehler (changes requested):

.idea/workspace.xml
szafran98: Done

szafran98:
Reopen for tests rerun

szafran98:
@andrewsmedina hey, it's looking the tests are finally passed

andrewsmedina:
@szafran98 thanks for contribute

PR #633, opened 2018-09-14
Adding support for buttons with a `form` attribute

theunraveler:
HTML5 allows for a form's submit button to live outside of the form itself, and instead specify a `form` attribute that points to the ID of the form that the button submits.

This PR fixes an error that occurred when clicking such a button.

PR #803, opened 2020-09-11
Adding support for buttons with a `form` attribute

theunraveler:
Reopen from https://github.com/cobrateam/splinter/pull/633, now with a test.


tests/form_elements.py
```diff
@@ -89,6 +89,13 @@ def test_clicking_submit_button_posts_button_value_if_value_present(self):
             "submit-button: submit-button-value",
         )
 
+    def test_submitting_a_form_when_the_button_is_outside_the_form(self):
+        self.browser.find_by_css('input[form="form-with-submit-outside"]').click()
+        self.assertEqual(
```
jsfehler, line 94: For new tests, it's better to use the pytest syntax:
```
assert "submit-input: submit-input-value" == self.browser.find_by_xpath("/descendant-or-self::*").text
```

theunraveler: Thanks, updated.

PR #820, opened 2020-10-07
Fixes #819, lxmldriver doing join the wrong way

### LINKED ISSUE #819: lxmldriver writing wrong address during POST when form action is empty
If you your current url is "/form" and you have this HTML:

```html
<form method="post">
   <button type="submit">send</button>
</form>
```

The form does not get "posted" to "/form" but to "/form/" and when you don't have trailing slashes in your URL paths you get 404.

The problem is here in the LXML driver:

```python
    def submit(self, form):
        import ipdb; ipdb.set_trace()
        method = form.attrib.get("method", "get").lower()
        action = form.attrib.get("action", "")  # <-- there is no action so this will be empty string
        if action.strip() != ".":  # <- if the URL is not "."
            url = os.path.join(self._url, action)  # <- this join returns "/form/", 
        else:
            url = self._url
        self._url = url
        data = self.serialize(form)

        self._do_method(method, url, data=data)
        return self._response
```

I think we should not be using `os.path.join` here because this is file system paths and is OS dependent, so windows you could get some backslashes. Use instead `urllib.parse.urljoin`:

```python
>>> from urllib.parse import urljoin
>>> urljoin('/form', '')
'/form'
```

edelvalle:


splinter/driver/lxmldriver.py
```diff
@@ -77,8 +81,8 @@ def serialize(self, form):
     def submit(self, form):
         method = form.attrib.get("method", "get").lower()
         action = form.attrib.get("action", "")
-        if action.strip() != ".":
-            url = os.path.join(self._url, action)
+        if action.strip() not in (".", ""):
+            url = urllib.parse.urljoin(self._url, action)
```
andrewsmedina, line 85: How the urljoin is imported, we should use it directly. This will fix the tests that are broken:

```suggestion
            url = urljoin(self._url, action)
```

andrewsmedina (changes requested):
Thanks for contributing. Can you add a test with the case that ensures that this new join is in the right way?

andrewsmedina:
@edelvalle can you sync your branch with the master?

jsfehler:
Closed in favour of #900

Issue #284, opened 2014-01-21
fill_form - option to ignore missing fields?


victorhooi:
Hi,

Currently, when you use `fill_form`, and there is an key present in your dict, that is not present on the page, it will raise a:

```
ElementDoesNotExist: no elements could be found with name "cvv2_cc"
```

Would it be possible to make configurable fill_form to ignore missing fields?

Basically, I'd like it to fill in as many of the fields as it can, and if there are any missing, just continue on. (In my case, I have a dict that I'd like to use on two different pages - one of those pages has an extra field, I'd rather not duplicate the whole dict again).

Thoughts?

Cheers,
Victor

jsfehler:
Implemented by #821

PR #821, opened 2020-10-11
Add flag to ignore missing fields in fill_form

jsfehler:
Implements #284


andrewsmedina:
@jsfehler can you add an update in the docs page about this new parameter?

andrewsmedina (approved):

Issue #164, opened 2012-08-27
support for external drivers


andrewsmedina:
It would be nice if splinter had a mechanism for register external drivers.

Issue #25, opened 2011-01-09
Support Internet Explorer


fsouza:
Add support to Internet Explorer using the webdriver Splinter driver.

parkouss:
Maybe appveyor can help here ? (http://www.appveyor.com/)
This is the windows equivalent for travis-ci, also free for open source projects.
Here is an example of the configuration file:
https://github.com/mozilla/mozregression/blob/master/appveyor.yml

kanth240:
For folks who want to use IE with splinter, check this out: https://stirunagari.wordpress.com/2017/08/20/using-internet-explorer-web-driver-with-splinter-framework/

ckolumbus:
just added a comment to the blog entry that @kanth240 mentioned above. with a minor tweak the solution there works perfectly for my scenario.
What is the issue that prevents us from integration this into splinter? Test env?

jsfehler:
Closing this issue. Internet Explorer is very, very dead at this point. Microsoft hasn't supported it in years. Splinter supports Edge but that's as far as we can go.

PR #841, opened 2020-11-04
Implement plugin system

jsfehler:
This solution uses Pluggy https://pluggy.readthedocs.io/en/latest/ to add a plugin system into Splinter.

This PR features hooks for:
- registering external drivers. This implements #164 and provides a solution for #25. Internet Explorer is an old browser at this point, but a custom driver could be dropped in without hacking splinter's internals.
- Performing actions after a page is navigated to
- Performing actions before closing the browser.

Plugins classes or modules can be registered via a call to `splinter.plugins.register()`

This PR is missing:
- Hooks implementation for non-webdriver drivers
- Documentation

But will be added if the overall approach works.

### DIFF
--- a/splinter/__init__.py
+++ b/splinter/__init__.py
@@ -4,5 +4,8 @@
 
 from splinter.browser import Browser  # NOQA
 
+from splinter.plugin_manager import hookimpl, plugins
 
 __version__ = "0.14.0"
+
+__all__ = ['hookimpl', 'plugins']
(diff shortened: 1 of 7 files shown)


jsfehler:
This is a bare bones example of what an external driver plugin could look like:

```from selenium.webdriver import Ie

import splinter
from splinter.driver.webdriver import BaseWebDriver


class WebDriver(BaseWebDriver):
    driver_name = "Ie"

    def __init__(self, wait_time=2):
        self.driver = Ie()

        super(WebDriver, self).__init__(wait_time)


class InternetExplorerPlugin:
    @splinter.hookimpl
    def splinter_prepare_drivers(drivers):
        drivers['internet_explorer'] = WebDriver


splinter.plugins.register(InternetExplorerPlugin())
```

PR #851, opened 2021-01-05
Extend user documentation for screenshot options.

mateoclarke:


jsfehler:
You're confusing the driver attribute of a Browser object and the DriverAPI base class that a Browser  inherits from. The driver attribute allows users to access underlying methods of the automation driver.

`save_screenshot` is a selenium method and already documented at: https://www.selenium.dev/selenium/docs/api/py/webdriver_remote/selenium.webdriver.remote.webdriver.html?highlight=save_screenshot#selenium.webdriver.remote.webdriver.WebDriver.save_screenshot

Issue #777, opened 2020-05-23
ElementAPI compatible with item getters


medecau:
Please, provide access to HTML attributes through item getters.

The following code makes three assertions of which only one works.
```
import operator
import splinter

b = splinter.Browser()
b.visit('https://www.google.com/')
q = b.find_by_name('q').first

assert 'q' == q.get('name')
assert 'q' == operator.itemgetter('name')(q)
assert 'q' == q['name']
```

_I would write the change myself but I have been unable to properly run the test-suite._

rg666:
Hi,
Is this issue still up for fix? I'd like to.

jsfehler:
@rg666 Go for it.

rg666:
Can you please assign to me.

rg666:
PR [856](https://github.com/cobrateam/splinter/pull/856)provided, would you please review it? Thank you! @jsfehler

jsfehler:
@rg666 @medecau Item getters appear to already work. I'll gladly accept unit tests for those behaviours since it appears they go untested at the moment.

I don't see any compelling reason to add `.get` as a method to ElementAPI

rg666:
Thank you for the review comments, it was my first PR. Ah, I was working on the review comments. If .get() method isn't required, please let me know. :) @jsfehler @medecau

jsfehler:
@rg666 I don't see any bugs involving item getters or missing functionality. I encourage you to write unit tests for the behaviour so we can verify they work. If the unit tests pass, I'll close this issue.

rg666:
Noted.

medecau:
> I don't see any compelling reason to add .get as a method to ElementAPI

Okay.

PR #856, opened 2021-02-12
Issue 777 fix

rg666:
Fix for issue [777](https://github.com/cobrateam/splinter/issues/777): Provided a  "get" method to access the HTML attributes' element list.

Test method test_method_get added to tests\test_element_list.py and tested successfully.

### DIFF
--- a/splinter/driver/webdriver/__init__.py
+++ b/splinter/driver/webdriver/__init__.py
@@ -748,6 +748,9 @@ class WebDriverElement(ElementAPI):
     def _get_value(self):
         return self["value"] or self._element.text
 
+    def get(self,value):
+        return self[value]
+    
     def _set_value(self, value):
         if self._element.get_attribute("type") != "file":
             self._element.clear()
--- a/tests/test_element_list.py
+++ b/tests/test_element_list.py
@@ -5,7 +5,8 @@
 # license that can be found in the LICENSE file.
 
 import unittest
-
+import operator
+import splinter
 from splinter.element_list import ElementList
 from splinter.exceptions import ElementDoesNotExist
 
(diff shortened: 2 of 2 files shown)


splinter/driver/webdriver/__init__.py
```diff
@@ -748,6 +748,9 @@ def __init__(self, element, parent):
     def _get_value(self):
         return self["value"] or self._element.text
 
+    def get(self,value):
```
jsfehler, line 751: This is incorrect. The implementation of getting with a list index must be done through `__getitem__`

tests/test_element_list.py
```diff
@@ -102,3 +103,15 @@ def test_not_found_exception_with_query_and_method(self):
         expected_message = 'no elements could be found with id "menu"'
         e = cm.exception
         self.assertEqual(expected_message, e.args[0])
+
+    def test_method_get(self):
+        """
+        should provide a \"get\" method
+        which provides access to HTML attributes' element list of the visited webpage
+        """
+        wd_ins = splinter.Browser()
```
jsfehler, line 112: You don't need a live browser instance to test getting attributes. You can just use a WebDriverElement instance.

jsfehler (changes requested):

splinter/driver/webdriver/__init__.py
jsfehler: To clarify:

I don't see a reason to add a `get` method.

However, if you were, you'd need to add it to the ElementAPI, which is the blueprint for WebDriverElement, and then support it across all the Elements, not just WebDriver.

Returning `self[value]` isn't what #777 was about. You would want to return the html attributes.

rg666: Valid point, I'll work towards the review points.

rg666:
Closing this PR.

PR #874, opened 2021-04-15
Updated doc for Mac os

caspii:
`brew cask install chromedriver` no longer works. See here https://github.com/Homebrew/discussions/discussions/902#discussioncomment-398348

Also, I just tried `brew install chromedriver` and that works too. No cask required!


jsfehler:
@caspii Thanks for the contribution.

PR #892, opened 2021-07-03
allow passing options to Firefox webdriver

aculich:
This change matches what's in chrome.py#L28 to allow passing in options when
creating a new Browser() window. Also renamed variable from `firefox_options` to
just `options` to make it generic to match chrome.py.

```
from selenium.webdriver import Firefox, FirefoxOptions
opts = FirefoxOptions()
opts.add_argument("--some-argument")
browser = Browser('firefox', options=opts)
```

### DIFF
--- a/splinter/driver/webdriver/firefox.py
+++ b/splinter/driver/webdriver/firefox.py
@@ -41,7 +41,7 @@ class WebDriver(BaseWebDriver):
         firefox_capabilities = DesiredCapabilities().FIREFOX
         firefox_capabilities["marionette"] = True
 
-        options = Options() if options is None else options
+        options = options or Options()
 
         if capabilities:
             for key, value in capabilities.items():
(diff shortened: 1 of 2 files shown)


splinter/driver/webdriver/firefox.py
```diff
@@ -40,7 +40,7 @@ def __init__(
         firefox_capabilities = DesiredCapabilities().FIREFOX
         firefox_capabilities["marionette"] = True
 
-        firefox_options = Options()
+        options = Options() if options is None else options
```
andrewsmedina, line 43: It is necessary to add the option as parameter in the `__init__` method definition.

And we can simplify this code using `or` boolean operator:
```suggestion
        options = options or Options()
```

andrewsmedina (changes requested):

aculich:
Added it to __init__() so thanks for pointing that out.

Also, simplified it in both the firefox and chrome drivers to just `options = options or Options()`

splinter/driver/webdriver/firefox.py
aculich: Add this simplification to both the firefox and chrome drivers.

Issue #929, opened 2021-10-13
Splinter doesn't work with Selenium 4.0


slafs:
Selenium 4.0.0 was just released.
With the latest version there are apparently some API changes that Splinter doesn't handle well.

Our issue was `TypeError: __init__() got an unexpected keyword argument 'timeout'` on https://github.com/cobrateam/splinter/blob/253a6f21699a60651943bac11cf87a24da4d5a81/splinter/driver/webdriver/firefox.py#L71

walksonair:
>  or upgrade to `splinter>=0.16.0` 😄

how do I do that? Whenever I run pup upgrade it keeps getting splinter==0.15.0

slafs:
> > or upgrade to `splinter>=0.16.0` 😄
> 
> how do I do that? Whenever I run pup upgrade it keeps getting splinter==0.15.0

you can try `pip install --upgrade splinter` or `pip install "splinter>=0.16.0"`

blaise-io:
> upgrade to `splinter>=0.16.0`

[0.16.0 pins `selenium<4`](https://github.com/cobrateam/splinter/blob/0.16.0/setup.py#L29) so that won't work.

This is unfortunate because [file uploads don't work on Selenium 3.x on Python 3.9](https://github.com/SeleniumHQ/selenium/issues/8762).

jsfehler:
@blaise-io 0.16.0's release pins selenium 3 in order to prevent dependency conflicts, it has nothing to do with getting selenium 4 to work.

Selenium 4 compatibility is coming soon-ish.

slafs:
> 0.16.0 pins selenium<4 so that won't work.

Right, the workaround I mentioned is for installing Splinter with Selenium 3.x. Not for making Splinter work with Selenium 4.x. Just to be clear.

jsfehler:
Fixed by #963

PR #930, opened 2021-10-13
Pin Selenium

slafs:
Selenium 4.0.0 was just released.
With the latest version there's apparently some API changes that Splinter doesn't handle well.

Our issue was `TypeError: __init__() got an unexpected keyword argument 'timeout'` on https://github.com/cobrateam/splinter/blob/253a6f21699a60651943bac11cf87a24da4d5a81/splinter/driver/webdriver/firefox.py#L71

As a workaround for now, until #929 gets properly fixed, we can probably pin the version of Selenium as a dependency to be lower than 4.0.


jsfehler:
@slafs Thanks for the contribution.

PR #970, opened 2021-12-20
`setup.py`: Add project URLs for PyPI

tony:
Makes it easier to open issues / change information

Example:
image


jsfehler:
@tony I'm good with this in general, but I'd prefer the changelog point to https://github.com/cobrateam/splinter/tree/master/docs/news
Unless you have a reason not to.

It probably makes more sense to migrate the individual files in news into one document called CHANGELOG.md. That would make linking to the changelog easier.

tony:
@jsfehler Updated URL, Rebased and pushed.

PR #982, opened 2022-01-17
Set firefox preferences through Options instead of FirefoxProfile

jsfehler:


andrewsmedina (approved):

splinter/driver/webdriver/firefox.py
```diff
@@ -5,10 +5,10 @@
 # license that can be found in the LICENSE file.
 
 from selenium.webdriver import Firefox
-from selenium.webdriver.firefox.firefox_profile import FirefoxProfile
-from splinter.driver.webdriver import BaseWebDriver
 from selenium.webdriver.firefox.options import Options
 
+from splinter.driver.webdriver import BaseWebDriver
```
blueyed, line 10: This should probably not get moved..

splinter/driver/webdriver/firefox.py
```diff
@@ -29,11 +29,11 @@ def __init__(
         **kwargs
     ):
 
-        firefox_profile = FirefoxProfile(profile)
-        firefox_profile.set_preference("extensions.logging.enabled", False)
-        firefox_profile.set_preference("network.dns.disableIPv6", False)
-
         options = options or Options()
+        if profile:
+            options.set_preference("profile", profile)
```
blueyed, line 34: In https://github.com/cobrateam/splinter/issues/1044#issuecomment-1154244607 I've used `options.profile` (after inspecting Selenium's code).

PR #1010, opened 2022-03-15
fix: element screenshots gets resized before cropping - cropping out too much of the image

### LINKED ISSUE #668: element screenshot code needs correction
In splinter ->driver->webdriver->__init__.py, line 799
###original code 
        ratio = im_width * 1.0 / window_width
        height_ratio = im_height / ratio

        im = im.resize((int(window_width), int(height_ratio)))

        location = self._element.location
        x, y = location['x'], location['y']

        pic_size = self._element.size
        w, h = pic_size['width'], pic_size['height']
        box = x, y, x + w, y + h
        #box = x, y, x + w, y + h
        box = [int(i) for i in box]
        target = im.crop(box)
 #############################
there was a by-ratio image processing that suppose to crop correct element screenshot from a window screen shot. Maybe there was history issues that lead to the fix. However , this do not work well when I am using up-to-date (chrome driver version 2.44,2.43,2.42, I did not try even earlier ones), cropped image is shifted toward right-bottom corner , in my case losing its' left 5 or 3 pixel lines and top 3 or 5 pixel line (depending on using full=True or not), and respectively the right and bottom lines.image size remains.
 
While debugged into code, found the driver captured window screenshot(aka the "im" in code is already the canvas part，no menu, no boundary.
Thus , just clean the processing code, use as-is crop get the perfect element screenshot, code replacing block above:
            
        location = self._element.location
        x, y = location['x'], location['y']

        pic_size = self._element.size
        w, h = pic_size['width'], pic_size['height']
        box = x, y, x + w, y + h
        box = [int(i) for i in box]
        target = im.crop(box)

python3 on win10 , 64bit.chrome driver

matslindh:
This may have been a legacy requirement, but gives the wrong results with both the Chrome and Firefox drivers currently. The screenshot gets scaled wider than necessary, leaving a slightly blurred and cut element.

This simply removes the resizing code as explained in #668.

Closes #668.

### DIFF
--- a/splinter/driver/webdriver/__init__.py
+++ b/splinter/driver/webdriver/__init__.py
@@ -1007,18 +1007,9 @@ class WebDriverElement(ElementAPI):
                                       'Please use "pip install Pillow" install it.')
 
         full_screen_png = self.driver.get_screenshot_as_png()
-
         full_screen_bytes = io.BytesIO(full_screen_png)
 
         im = Image.open(full_screen_bytes)
-        im_width, im_height = im.size[0], im.size[1]
-        window_size = self.driver.get_window_size()
-        window_width = window_size['width']
-
-        ratio = im_width * 1.0 / window_width
-        height_ratio = im_height / ratio
-
-        im = im.resize((int(window_width), int(height_ratio)))
 
         location = self._element.location
         x, y = location['x'], location['y']


jsfehler:
@matslindh Thanks for the contribution. I've gone with a different route with this fix in #1020

PR #1029, opened 2022-04-16
Added data collector automation tool

Pisciotta:
The data collection is meant to provide an easy way to collect information over different pages whose navigation is automated and can support user interactions (e.g. wait for user login and afterwards visit some sub-urls and for each of them extract some information). Class implementation (splinter/data_collector.py), usage (docs/data-collector.rst) and test cases (tests/test_data_collector.py) are provided.

### DIFF
--- a/docs/data-collector.rst
+++ b/docs/data-collector.rst
@@ -1,0 +1,62 @@
+.. Copyright 2012 splinter authors. All rights reserved.
+   Use of this source code is governed by a BSD-style
+   license that can be found in the LICENSE file.
+
+.. meta::
+    :description: Splinter data collection automation tool
+    :keywords: splinter, python, data collection
+
++++++++++++++++++
+Data Collector
++++++++++++++++++
+
+Splinter can be used to simplify data collection automation.
+
+The following example shows how to use the DataCollector class.
+
+Gathering Twitter profiles data after login
+=========================
+
+This script will first visit Twitter login page (using Chrom in incognito mode (#1) 
+and waits until the user manually enter his credentials i.e. for the twitter home
+url to be automatically visited on successful loin (#2)
+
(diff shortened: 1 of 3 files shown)


jsfehler:
Sorry, but I'm going to reject this PR for several reasons.

- This code runs parallel to splinter. It doesn't implement or extend splinter's API, it takes an instance of a browser.

- It implements new features that are non-orthogonal. For example, the `wait_until_url` method is isolated to the DataCollector class for no reason and the callback system relies on private methods inside the DataCollector class.

- It only uses splinter to extract HTML from a page. Splinter is a consistent API on top of various other libraries designed for interacting with web sites.

- A regex is used to parse a page's html content as text instead of using the browser's API to interact with elements.

- The `wait_until_url` method implements a hard-coded sleep. This is far too flaky to risk being used.

- A method like `add_url_template` is pointless in python. Lists already have an append method. Creating a wrapper method won't guard against anything but it will make the code less maintainable.

- Errors are hidden by the logic of the `iterate_over_pages`. If a placeholder is not in the URL templates, it's silently ignored.

- Custom string templating logic is used instead of the standard library.

- The docstring is incorrectly formatted.

- The unit tests call a third party URL.

- The unit tests are written with unittest instead of pytest.

Most importantly of all, I fail to see the fundamental purpose of this class. It restricts built-in functionality and can easily be duplicated without any new features or restrictions. An example would be:

```
data_collection = {}

pages = ["tldrnewsletter", "cnni", "realpython", "elonmusk"]
url_template = "https://twitter.com/{page}"

browser.visit("https://twitter.com/home")
for page in pages:
    browser.visit(url_template.format(page=page))
    elements = browser.find_by_tag('span')

    data = []    
    for element in elements:
        if "Joined" in element.value:
            data.append(element.value)
            data_collection[f"{page}"] = data
```

PR #1048, opened 2022-06-10
github/workflows: add workflow for uploading to PyPI

fsouza:
This will happen automatically when we tag the repo.

TODO:

- [ ] Setup a token (preferably with a "bot" account, but I'm happy to generate a splinter-only token with my account)

### DIFF
--- a/.github/workflows/upload-to-pypi.yml
+++ b/.github/workflows/upload-to-pypi.yml
@@ -1,0 +1,26 @@
+name: Upload to PyPI
+on:
+  push:
+    tags:
+      - "*"
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


fsouza:
@jsfehler @andrewsmedina let me know what yall think

jsfehler:
@fsouza This is cool, it will make releases much less painful for everyone. As long as the pypi token isn't anyone's personal token,  let's merge it.

fsouza:
> @fsouza This is cool, it will make releases much less painful for everyone. As long as the pypi token isn't anyone's personal token,  let's merge it.

Sounds good. I'm AFK right now, but later I'll generate a project token for splinter, configure the secret here and merge this PR.

I'll also do a test with `--skip-existing` before merging, just to make sure this works end to end.

PR #1049, opened 2022-06-11
github/workflows/upload-to-pypi: install wheel

fsouza:
Should fix the workflow.


fsouza:
It works: https://github.com/cobrateam/splinter/runs/6846405335

Issue #1052, opened 2022-06-16
Unnecessary requirement of selenium with `from splinter import Browser`


blueyed:
When using `from splinter import Browser` to only use e.g. `Browser("django")`, it fails without selenium being installed due to: https://github.com//cobrateam/splinter/blob/986ce0a10c52f08196b32b91f752182cb7517892/splinter/browser.py#L14

This is only used within https://github.com//cobrateam/splinter/blob/986ce0a10c52f08196b32b91f752182cb7517892/splinter/browser.py#L73-L85, and could be handled through some conditional import, i.e. without selenium being installed/importable, this exception would not happen - and more specifically is only expected to happen for a certain set of the browsers (`splinter.driver.webdriver.*`).

It would be great if the wrapper could be made to not require installation of selenium for non-webdriver-based clients.

PR #1053, opened 2022-06-16
[RFC] Do not require Selenium to be installed only for WebDriverException

blueyed:
Fixes https://github.com/cobrateam/splinter/issues/1052

### DIFF
--- a/splinter/browser.py
+++ b/splinter/browser.py
@@ -13,6 +13,13 @@ from urllib3.exceptions import MaxRetryError
 
 from splinter.exceptions import DriverNotFoundError
 
+driver_exceptions = (IOError, HTTPException, MaxRetryError)
+try:
+    from selenium.common.exceptions import WebDriverException
+    driver_exceptions += (WebDriverException,)
+except ImportError as e:
+    logging.debug(f"Import Warning: {e}")
+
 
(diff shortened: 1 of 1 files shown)


splinter/browser.py
```diff
@@ -78,12 +76,24 @@ def get_driver(driver, retry_count=3, *args, **kwargs):
 
     """
     err = None
+    handle_excs = ()
```
jsfehler, line 79: @blueyed I think a better approach here is to do the try/except for the exception classes outside the browser init loop:

```
driver_exceptions = (IOError, HTTPException, MaxRetryError)
try:
    from selenium.common.exceptions import WebDriverException
    driver_exceptions += (WebDriverException,)
except ImportError as e:
    logger.debug(f"Import Warning: {e}")
```

and then:
```
except driver_exceptions as e:
    err = e
```

jsfehler (changes requested):

splinter/browser.py
blueyed: Makes sense, pushed a fixup (which should get squash-merged).

However, please note that my initial idea was to delay the import until it is necessary (in case of an error), but I am fine with having it being tried always.

PR #1054, opened 2022-06-17
Use named logger instead of root logger

jordanvance:
use named logger for splinter

closes #1054

### DIFF
--- a/splinter/browser.py
+++ b/splinter/browser.py
@@ -26,6 +26,8 @@ _DRIVERS = {
     'zope.testbrowser': None,
 }
 
+LOG = logging.getLogger(__name__)
+
 try:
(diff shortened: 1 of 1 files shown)


splinter/browser.py
```diff
@@ -26,6 +26,8 @@
     'zope.testbrowser': None,
 }
 
+LOG = logging.getLogger(__name__)
```
fsouza, line 29: nit: can we call the variable `logger` instead of `LOG`?

fsouza (commented):
Thanks!

splinter/browser.py
jordanvance: certainly

PR #1066, opened 2022-07-07
Cleanup type hints

jsfehler:


splinter/browser.py
```diff
@@ -24,7 +26,7 @@
     logger.debug(f"Import Warning: {e}")
 
 
-_DRIVERS = {
+_DRIVERS: Dict[str, Union[None, Type[DriverAPI]]] = {
```
fsouza, line 29: any thoughts on using `from __future__ import annotations`? Then you could rewrite this as:

```suggestion
_DRIVERS: dict[str, type[DriverAPI] | None] = {
```

Basically we'd be able to:

- replace `typing.Dict` with `dict`
- replace `typing.Tuple` with `tuple`
- replace `typing.Union[A, B]` with `A | B`
- replace `typing.Optional[A]` with `A | None`

Assuming Python 3.6 is no longer supported, that should be fine. We could even use pyupgrade to automatically rewrite those.

jsfehler: This is a good idea, generally speaking, but I have a feeling this could break when mixed with modifying the InheritedDocs class to inherit annotations as well as docstrings. I'll be tinkering with that soonish and if it doesn't break I'll make these changes to the type hints.

Issue #1080, opened 2022-08-05
attach_file work with flask?


marcuxyz:
Hi,
I have a question.

`browser.attach_file` work with flask?
I've been trying to work with this for almost 1 month. :(

an error always happens when the redirect is done:
`ValueError: read of closed file `

Maybe... if do test with `return redirect` instead `return "|".join(buffer) `, this break:

https://github.com/cobrateam/splinter/blob/bd7b82161802f07a7f79e6b1e74557f4cf23d029/tests/fake_webapp.py#L117

I change the code to:

```python
@app.route("/upload", methods=["GET", "POST"])
def upload_file():
    if request.method == "POST":
        f = request.files["file"]
        return redirect('/')
```

and run the test, with the follow command:

```shell
tox -e tests -- tests/test_flaskclient.py
```

<img width="1714" alt="Screen Shot 2022-08-05 at 11 12 45" src="https://user-images.githubusercontent.com/9499562/183095325-a785096b-1427-4bdb-b18d-396a8b72e847.png">

Thanks

jsfehler:
@marcuxyz Can you provide the command you're using to run the tests, a full traceback, plus the operating system and python version you're using?

marcuxyz:
@jsfehler I'm use python 3.10 and Mac OS Monterey.  `pytest` is the command used to run tests.

PR #1083, opened 2022-08-12
Fix attach_file method to FlaskClient

marcuxyz:
I recently opened an issue #1080 . When we try to redirect the user to another page, after having sent a file through the form. The following error occurs:

```
ValueError: read of closed file
```

This happens, because, to redirect user for another page using some test client is necessary  ensure that to `GET` the `data` object is not sent.


fsouza (approved):
That looks correct to me.

@jsfehler any thoughts?

PR #1125, opened 2023-01-18
Remove selenium 3 support

jsfehler:
The Selenium 3.x series is very dead at this point. For a while it made sense to support it because:

- Selenium 4.0 had deprecation warnings as opposed to removals.
- Splinter did not have to write very much custom code to handle both 3.x and 4.x

Selenium 4.x is moving past these deprecation warnings and it's not worth it to maintain both as a back-end for Splinter as Selenium 4.x continues diverging. At this point, anyone who absolutely needs Selenium 3.x will have to make do with Splinter 0.19.0.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -47,62 +47,13 @@ jobs:
 
       - name: Run tests
         run: |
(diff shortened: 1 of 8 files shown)


fsouza (approved):
nice

PR #1131, opened 2023-01-18
Increase max selenium version allowed

jsfehler:


setup.py
```diff
@@ -57,7 +57,7 @@ def get_version_data() -> dict:
         "zope.testbrowser": ["zope.testbrowser>=5.5.1", "lxml>=4.2.4", "cssselect"],
         "django": ["Django>=2.0.6", "lxml>=4.2.4", "cssselect"],
         "flask": ["Flask>=2.0.2", "lxml>=4.2.4", "cssselect"],
-        "selenium": ["selenium>=4.1.0,<4.3.0"],
+        "selenium": ["selenium>=4.1.0,<4.5.0"],
```
blueyed, line 60: What's the rationale here?  Could it be `<4.9.0` already, given that the current release is 4.8?

What do you think about a matrix on CI with different Selenium versions?

jsfehler: When Selenium is installed separately from Splinter, you can use whichever version you can make work. When Selenium is installed as an extra of Splinter, we need to ensure the dependency works with the released version. At some point after 4.5.0, Selenium and/or geckodriver made breaking changes as to when an error message is raised. Some of Splinter's tests caught this. The test on our side could be better but there's still a change in behaviour that needs to be accounted for and I haven't had time to investigate.

PR #1134, opened 2023-01-24
Add splinter.Config, deprecate many browser arguments

### LINKED ISSUE #602: Headless mode in remote webdriver
Not sure if its possible, but it would be great if we can look into getting headless mode working via remote webdriver.

Referencing discussion on https://github.com/cobrateam/splinter/issues/597#issuecomment-377111031

jsfehler:
Fixes #602

This PR adds a new object called Config. It's designed to act as a simple alternative to Selenium Options objects and replace custom arguments to the various drivers. It also allows Remote Webdriver to benefit from the same behaviour as the other Selenium drivers.

e.g.: Instead of:
```
Browser('chrome', headless=True, fullscreen=True)
```

It's now:

```
my_config = Config(headless=True, fullscreen=True)
Browser('chrome', config=my_config)
```

This reduces the number of explicit arguments sent to the browser and still allows custom arguments to be sent directly to the driver.

### DIFF
--- a/splinter/driver/webdriver/setup.py
+++ b/splinter/driver/webdriver/setup.py
@@ -1,0 +1,95 @@
+from selenium.webdriver import Remote
+
+"""The following functions prepare Webdriver for use.
+
+Config, Options, Service, and other arguments are configured and sent to
+the driver.
+
+The driver is returned.
+"""
+
+
+def _setup_chrome(driver_class, config=None, options=None, service=None, **kwargs):
+    """
+    Returns: selenium.webdriver.Chrome || selenium.webdriver.Remote
+    """
+    if config.user_agent is not None:
+        options.add_argument(f"--user-agent={config.user_agent}")
+
+    if config.incognito:
+        options.add_argument("--incognito")
+
+    if config.fullscreen:
+        options.add_argument("--kiosk")
+
+    if config.headless:
+        options.add_argument("--headless")
+        options.add_argument("--disable-gpu")
+
+    if config.extensions:
+        for extension in config.extensions:
+            options.add_extension(extension)
+
+    if driver_class == Remote:
+        rv = driver_class(options=options, **kwargs)
+    else:
+        rv = driver_class(options=options, service=service, **kwargs)
+
+    return rv
+
+
+def _setup_edge(driver_class, config=None, options=None, service=None, **kwargs):
+    """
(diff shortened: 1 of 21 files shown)


splinter/driver/webdriver/setup.py
```diff
@@ -0,0 +1,95 @@
...
+"""The following functions prepare Webdriver for use.
+
+Config, Options, Service, and other arguments are configured and sent to
+the driver.
+
+The driver is returned.
+"""
+
+
+def _setup_chrome(driver_class, config=None, options=None, service=None, **kwargs):
```
fsouza, line 12: Do these need to default to `None`? we're always providing it, no?

fsouza (approved):
Nice, I like this. Just one question on the parameters of the setup functions, non-blocking though since they're private API anyways.

splinter/driver/webdriver/setup.py
jsfehler: I tend to prefer keyword arguments with None as a default to positional arguments. It's just personal preference, it doesn't make a difference for a private function like this.

Because this moves almost all the logic out of the Webdriver sub-classes, there's probably a more evolved version of these functions that have a closer relationship to the Config and/or Browser objects. If there's no reason to sub-class splinter's webdriver class, we'll need to build an interface for handling Config to driver communication if we want to expose any of this to the public.

fsouza: Oh, the reason this caught my attention is that the code in the function clearly assumes that `config` is never `None` and that `options` should probably never be `None`, but the function signature indicates that they are usually `None`. It's not just a matter of API, but rather making it clear to the reader what they can expect from this function without having to dig into details. When I see a default value, I expect the code to work with it.

jsfehler: That's not wrong, but a None value will cause an error where it fails whereas a missing positional argument will cause a TypeError on the function call. I find the latter less informative because it only tells you that something is missing. The former lets you know that something is missing and something downstream is exploding as a result. I treat it on a situation-by-situation basis. If I'm writing something as an interface, I'd err on the side of the TypeError on the function call. I don't mind changing it.

fsouza: A good static analysis tool should be able to catch the missing argument, no need to wait for a runtime exception. I definitely don't think we need to block on this though, it's a private API and we can change it at anytime. Feel free to ship!

Issue #1173, opened 2023-06-08
Exception when using the latest Selenium


mpibpc-mroose:
Hi,
for my tests I always use the last Selenium docker image:

```
version: '3.7'

services:
  django:
    build:
      context: .
    depends_on:
      - selenium

  selenium:
    image: selenium/standalone-chrome:latest
    shm_size: '4G'
```

With the last Selenium I get the following exception:

```
Traceback (most recent call last):
  File "/source/lib/unittest.py", line 49, in setUpClass
    cls.browser = Browser(
                  ^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/splinter/browser.py", line 125, in Browser
    return get_driver(driver, retry_count=retry_count, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/splinter/browser.py", line 93, in get_driver
    return driver(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/splinter/driver/webdriver/remote.py", line 44, in __init__
    driver = Remote(command_executor, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: WebDriver.__init__() got an unexpected keyword argument 'desired_capabilities'
```

- `Digest: sha256:da6d0894a465c938a8134f44c1bfcc4e0adaca45296dc65d3fed87e1b6b8ee2f`was fine
- `Digest: sha256:9fd0fe4f8a39f65aeecbdb38fc35eda6e505f217d9c0f0d3420b17cd187e9216` fails

jsfehler:
Fixed by #1202

PR #1202, opened 2023-10-08
moved desired_capabilities inside if statement to add compatibility …

webjunkie01:
According to this commit here https://github.com/SeleniumHQ/selenium/commit/9f5801c82fb3be3d5850707c46c3f8176e3ccd8e#diff-89ba579445647535b74423c7bf4b8be79ef1ce33847a2768e623c3083a33545d

`desired_capabilities` has been removed from selenium 4.10.0 and up. This commits allows for old and new versions of selenium to work either using the newer `options` argument or the old `desired_capabilities`. This also prevents the error from selenium `TypeError: WebDriver.__init__() got an unexpected keyword argument 'desired_capabilities'`
It fixes issue [#1173](https://github.com/cobrateam/splinter/issues/1173)

### DIFF
--- a/splinter/driver/webdriver/remote.py
+++ b/splinter/driver/webdriver/remote.py
@@ -43,7 +43,7 @@ class WebDriver(BaseWebDriver):
             # Combine user's desired capabilities with default
             caps.update(kwargs["desired_capabilities"])
 
-        kwargs["desired_capabilities"] = caps
+            kwargs["desired_capabilities"] = caps
 
         kwargs["command_executor"] = command_executor
 


brondsem:
IMO this is good.  It fixes https://github.com/cobrateam/splinter/issues/1173 which would be really good to resolve.  Anyone able to review and merge this?

splinter/driver/webdriver/remote.py
```diff
@@ -43,7 +43,7 @@ def __init__(
             # Combine user's desired capabilities with default
             caps.update(kwargs["desired_capabilities"])
 
-        kwargs["desired_capabilities"] = caps
+            kwargs["desired_capabilities"] = caps
```
fsouza, line 46: Isn't this a change of behavior? What's the impact of no longer setting desired_capabilities to `{}`?

jordanvance: @fsouza `desired_capabilities` was deprecated; it was fully removed as part of selenium 4.10 so now you would get an invalid/unexpected keyword argument. Options that were passed as desired_capabilities are now handled with other method calls

fsouza: Right, but what's the impact of this change on code running with selenium <4.10?

webjunkie01: Would it make more sense to do a version check of the imported selenium lib and then either keep or delete `desired_capabilities` accordingly?

fsouza: Ohh I see that the parameter has been deprecated for ages. This PR looks good then.

fsouza (approved):

jordanvance:
@fsouza if it's possible to get a new release of splinter with this patch in it, that'd be amazing

fsouza:
Sounds good, I'll prep that release later this week if no one beats me to it

cc @jsfehler

PR #1238, opened 2024-02-14
Migrate from setup.py to pyproject.toml

jsfehler:
I've chosen hatchling over setuptools as I believe it offers a good value proposition for Python projects that don't require external modules or compilation of dependencies. For more information: https://hatch.pypa.io/dev/why/

This doesn't prevent us from using the `build` package as a front-end. Using `Hatch` as a replacement for `tox` and `build` could be done in the future, but isn't necessary immediately.


fsouza (approved):
Awesome, thanks for making this change!

PR #1240, opened 2024-02-14
Remove makefile, add build_docs command to tox.ini

jsfehler:
The makefile is super broken and has been for a long time. Since we use tox there's no reason to keep the makefile.


fsouza (approved):
Nice! 🎉

PR #1251, opened 2024-02-23
Start using a generic retry function

jsfehler:
- Rewrite `_find` to only handle the Selenium actions.
- Rewrite `find_by` to use the generic `_retry` function.
- Rewrite the `_retry` function to return the result of the function instead of a bool.


fsouza (approved):
Nice!

PR #1254, opened 2024-02-27
Use ABC for CookieManagerAPI

jsfehler:
This PR:
  - Adds the `__contains__` method to the CookieManagerAPI interface
  - Uses the ABC module to enforce the Splinter API 

Using ABC as a base class has some advantages for creating interfaces. The abstractmethod decorator forces subclasses to implement the decorated methods. Eventually, I'd like to have DriverAPI and ElementAPI use ABC as a base class.


fsouza (approved):
Sorry, somehow I missed this. LGTM

Issue #1147, opened 2023-03-19
Cannot send characters directly to the browser


LewisW:
The documentation implies that [you can send keys directly to the browser](https://splinter.readthedocs.io/en/latest/selenium-keys.html), but the [code expects an element name to be provided](https://github.com/cobrateam/splinter/blob/265cd9bd06645cfb1ba6e9393af4af8087421e99/splinter/driver/__init__.py#L272). 

The result is an error like the following:
`TypeError: BaseWebDriver.type() missing 1 required positional argument: 'value'`

Is this something that's still supported?

nameloCmaS:
From what I understand, the way to do this is as below.

```
from selenium.webdriver.common.keys import Keys
from splinter import Browser
from splinter.driver import webdriver

browser = Browser("firefox")
browser.visit("https://github.com")
driver = browser.driver
actions = webdriver.ActionChains(driver)
# Queue for effect only. The first / brings up the search bar,
# the others type the search query in the search bar, followed by
# Enter to submit the form.
send_keys = ["/", "cobrateam", "/", "splinter", Keys.ENTER]
for send in send_keys:
    actions.send_keys(send)

actions.perform()

```

The function ([implementation for web drivers](https://github.com/cobrateam/splinter/blob/265cd9bd06645cfb1ba6e9393af4af8087421e99/splinter/driver/webdriver/__init__.py#L571)) _could_ be edited to run this code if no `name` argument is provided. But the function signature would not allow a default to just be set for `name` without `value` also having a default. Changing this would be a breaking change and affect some implementations.


[More info](https://stackoverflow.com/a/28482440/3822101) from Stack Overflow.


Thus I agree, the documentation is misleading re this.

PR #1255, opened 2024-02-28
Tidy up after reading contributing docs

nameloCmaS:
This pull request does the below:

1. Tidies the contribute docs generally (updated some out of date links, added some more detail to code examples, updated RST syntax to use code-block in place of highlight descriptor, corrected tox test environments).
2. Changed Firefox example path from canary to firefox_nightly.
3. Removed old rst file referencing install of PyQt4 for now unused driver.
4.  Removed link to non-existent rst in old releases.
5. Updated Selenium keys as example code does not work as mentioned in issue #1147.
6. Corrected tutorial to ref second btnK named element as didn't work.
7. Added missing lint environment to tox.ini

For item 7 I have rewritten the tox.ini file but will raise an issue for bigger changes (#1256).


splinter/driver/webdriver/__init__.py
```diff
@@ -97,21 +97,17 @@ def next(self):  # NOQA: A003
         next_handle = self._browser.driver.window_handles[next_index]
         return Window(self._browser, next_handle)
 
-    def is_current():
```
jsfehler, line 100: Using decorators here is good, but I'd prefer if you split them into a separate PR. They're not related to the rest of the changes.

docs/contribute/contribute.rst
```diff
@@ -103,11 +109,13 @@ Build
 The `build_docs` environment is a wrapper around Sphinx's Makefile.
 Arguments will be passed to the Makefile. Thus, to build the docs in HTML format:
 
-.. highlight:: bash
+.. code-block:: bash
+
+  tox -e build_docs -- html
```
jsfehler, line 114: Thanks for catching this, typo on my part.

jsfehler (changes requested):

splinter/driver/webdriver/__init__.py
nameloCmaS: Reverted file; will submit separately.

nameloCmaS (commented):
Removed change to code as requested.

jsfehler (approved):

